PR #704, opened 2022-05-30
create spatial enrichment script

LBHSBALLEY:
Pyspark script to add wards etc. to point data and create lat/lon or eastings/northings if needed. Not used in any pipeline yet. Can be tested in Glue console with the job of the same name.


annajgibson (commented):
Would it be possible to tweak the inputs/outputs so that tables from different source data catalogues can be read? We have a req from Environmental Services to do some point-in-area calculations.

PR #796, opened 2022-07-11
Update main.py

daroclark:
Change the days returned from 1 to 7 to overcome missing days of ingestion.


LBHTKarki (commented):
Would be good to identify where the issues are and then see if we can resolve them without extending the timeframe first.

ryanbratten (approved):

daroclark:
Issue superseded by watermarking

PR #819, opened 2022-07-20
DPP-188 update lambda to trigger a crawler instead of a job

LBHTKarki:
[DPP-188](https://hackney.atlassian.net/browse/DPP-188)

There's an issue with the current setup where the lambda starts a glue job directly instead of starting a trigger. Because of this the crawler never runs.

This PR updates the flow so that the lambda starts the trigger which in turn starts the glue job. This ensures all consecutive steps get triggered as expected.

- update the lambda to take the trigger name as input parameter rather than the glue job name
- some refactoring for the lambda for easier testing
- update the api ingestion module to use the trigger name instead of the job name 
- add new optional input for the api ingestion lambda module
- update the lambda's permissions so it can start the trigger
- add new output value to the glue job module to return the trigger name

This setup has been tested in dev enviroment.


ryanbratten (approved):

LBHBReynolds-Carr (approved):

LBHTKarki:
> @LBHTKarki code wise this looks good but can you check with @LBHBReynolds-Carr that the changes to the api-ingestion module work with the SFTP stuff Bens been working on?

@ryanbratten just checked with @LBHBReynolds-Carr and there hasn't been any work done on the SFTP ingestion just yet and iCaseWork is the only operational ingestion process that's using this new more generic module. There might be some further tweaks we need to make later once the SFTP work kicks off, but this should be fine for now.

ryanbratten (approved):

PR #876, opened 2022-08-23
WIP:DPP-197 - Secondary CIDR blocks for each VPC - DO NOT MERGE

ryanbratten:


ryanbratten:
I have made this change on Dev successfully and then ran a terraform plan against the core and etl areas. ETL showed no changes to be made however in core it showed that the kafka cluster must be recreated which will be problematic. After talking to Matt K we've agreed to have a call with Matt, Tuomo and Tomek to see if there is a better way of implementing this fix

PR #896, opened 2022-09-06

barnesm707:
Second attempt as I had created an incorrect .csv for Ringgo daily transactions


barnesm707:
Second update for 06/09/2022 this should be process last as the previous github update was incorrect

PR #917, opened 2022-09-21
DPP-286 Add prevent destroy to glue catalog database resources

LBHTKarki:
## Link to JIRA ticket
[DPP-286](https://hackney.atlassian.net/browse/DPP-286)

### *What is the problem we're trying to solve*

We have Glue catalog databases that have been deployed using Terraform. If for any reason Terraform detected a change in configuration that would result in those resources to be redeployed and therefore deleted we woud lose all tables and data in those databases. Some of that data could potentially be difficult and time consuming to recover. 

### *What changes have we introduced*

Terraform has a resource lifecycle management feature that can be used to prevent resources from being destroyed. 

When the below is added to a resource any Terraform plan that would cause this particular resources to be destroyed will fail and therefore prevents the resource from being deleted. 

`lifecycle {
    prevent_destroy = true
  }`

### *How Has This Been Tested?*

Please see the Jira ticket for test plan and results

#### _Checklist_

- [ ] Added end-to-end (i.e. integration) tests where necessary e.g to test complete functionality of a newly added feature
- [ ] Added tests to cover all new production code
- [ ] Added comments to the README or updated relevant documentation _(add link to documentation)_, where necessary.
- [ ] Checked all code for possible refactoring
- [x] Code pipeline builds correctly

### *Follow up actions after merging PR*

N/A

### DIFF
--- a/terraform/core/03-input-derived.tf
+++ b/terraform/core/03-input-derived.tf
@@ -49,4 +49,8 @@ data "aws_ssm_parameter" "aws_vpc_id" {
 // ==== LANDING ZONE ===========
 resource "aws_glue_catalog_database" "landing_zone_catalog_database" {
   name = "${local.identifier_prefix}-landing-zone-database"
+
+  lifecycle {
+    prevent_destroy = true
+  }
 }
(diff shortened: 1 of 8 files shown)


timburke-hackit:
do these apply to dev too? and does that potentially create a problem with not being able to destroy the databases there?

LBHTKarki:
> do these apply to dev too? and does that potentially create a problem with not being able to destroy the databases there?

That's a good point and will stop you from running destroy on dev. For now we can manually update the values to false when running destroy there, but it's worth looking into something that would handle it better.

timburke-hackit (approved):

PR #937, opened 2022-10-04
DPP-295 Add copy from landing to raw zone for ringgo

LBHTKarki:
These updates complete the initial test setup for RingGo SFTP ingestion.

- Add new Glue Catalog database just for the SFTP data to keep it separate for now
- Add new job that copies daily CSV files from `/ringgo/sftp/input` landing zone folder to a folder in the raw zone. The job is using the existing CSV to raw script used elsewhere in parking to avoid duplication of code. These can be separated in the future if there's need to make them work differently
- Update the ingestion job Lambda name to be unique
- Update the target folder in the ingestion module to match the folder structure on the landing zone  

Please note this is pre-prod setup only and subject to change based on UAT.


timburke-hackit (approved):
lgtm

PR #973, opened 2022-10-27
DPP-265 run s3 file transfer on glue job success

LBHTKarki:
Update RentSense file transfer to trigger on Glue job (Rentsense outputs) status change (SUCCEEDED) rather that on set schedule.

This will be active on per-prod and prod.


timburke-hackit (approved):

LBHTKarki:
job name is not correct for each environment, so closing this and will create a new PR with correct config

PR #974, opened 2022-10-27
DPP-265 trigger rentsense transfer onm glue job success

LBHTKarki:
Update RentSense file transfer to trigger on Glue job (Rentsense outputs) status change (SUCCEEDED) rather that on set schedule.

This will be active on per-prod and prod.


LBHSBALLEY (approved):

LBHTKarki:
Closing because the config was missing some deployment rules. New PR will be created with further changes

PR #992, opened 2022-11-07
Dpp 315 alloy

timburke-hackit:
In order to handle non-UTF-8 characters present in some of the exported csv files the `alloy_api_export.py` script now
- Reads imported csv into a spark dataframe
- writes the dataframe as parquet to the raw zone, in UTF-8 encoding

Updated the `alloy_export_crawler` to crawl the parquet files created above.

Updated [alloy_raw_to_refined.py](https://github.com/LBHackney-IT/Data-Platform/compare/dpp-315-alloy?expand=1#diff-2ffa8a3301f6316127f0863742697cdb6fb9d4e5e487f6246421c95f8b83890f) so that it now creates partitions for the date the files are processed

Disables an unnecessary trigger in pre-production.

Add tests for create_s3_key()
Tested full pipeline on affected files in pre.

### DIFF
--- a/scripts/jobs/env_services/alloy_api_export.py
+++ b/scripts/jobs/env_services/alloy_api_export.py
@@ -14,7 +14,14 @@ from awsglue.job import Job
 from awsglue.transforms import *
 from awsglue.utils import getResolvedOptions
 from pyspark.context import SparkContext
-from scripts.helpers.helpers import get_glue_env_var, get_secret
+
+from scripts.helpers.helpers import (
+    PARTITION_KEYS,
+    add_import_time_columns,
+    clean_column_names,
+    get_glue_env_var,
+    get_secret,
+)
 
 
 def api_response_json(response):
@@ -32,7 +39,7 @@ def api_response_json(response):
     return json.loads(response.text)
 
 
-def create_s3_key(s3_downloads_prefix, import_date, file, prefix_to_remove=None):
+def create_s3_key(s3_prefix, file, prefix_to_remove=None, import_date=False):
     """
     creates the key argument for saving output files to s3
     """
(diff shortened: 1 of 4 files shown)


scripts/jobs/env_services/alloy_api_export.py
```diff
@@ -41,17 +50,26 @@ def create_s3_key(s3_downloads_prefix, import_date, file, prefix_to_remove=None)
...
-        pre = pre.lower()
+    if prefix_to_remove == None:
+        pass
+    else:
+        pre = prefix_to_remove.lower()
+
         if not file_table_name.startswith(pre):
             pass
         else:
-            file_table_name = file_table_name[len(pre):]
```
timburke-hackit, line 49: > PEP 8 recommends to treat : in slices as a binary operator with the lowest priority, and to leave an equal amount of space on either side, **except if a parameter is omitted (e.g. ham[1 + 1 :])**. It recommends no spaces around : operators for “simple expressions” (ham[lower:upper]), and **extra space for “complex expressions” (ham[lower : upper + offset])**. **Black treats anything more than variable names as “complex” (ham[lower : upper + 1]).** It also states that for extended slices, both : operators have to have the same amount of spacing, except if a parameter is omitted (ham[1 + 1 ::]). Black enforces these rules consistently.

> This behaviour may raise E203 whitespace before ':' warnings in style guide enforcement tools like Flake8. **Since E203 is not PEP 8 compliant, you should tell Flake8 to ignore these warnings**.

scripts/jobs/env_services/alloy_api_export.py
```diff
@@ -41,17 +50,26 @@ def create_s3_key(s3_downloads_prefix, import_date, file, prefix_to_remove=None)
...
+            file_table_name = file_table_name[len(pre) :]
```
timburke-hackit, line 61: > PEP 8 recommends to treat : in slices as a binary operator with the lowest priority, and to leave an equal amount of space on either side, **except if a parameter is omitted (e.g. ham[1 + 1 :])**. It recommends no spaces around : operators for “simple expressions” (ham[lower:upper]), and **extra space for “complex expressions” (ham[lower : upper + offset])**. **Black treats anything more than variable names as “complex” (ham[lower : upper + 1]).** It also states that for extended slices, both : operators have to have the same amount of spacing, except if a parameter is omitted (ham[1 + 1 ::]). Black enforces these rules consistently.

> This behaviour may raise E203 whitespace before ':' warnings in style guide enforcement tools like Flake8. **Since E203 is not PEP 8 compliant, you should tell Flake8 to ignore these warnings**.

LBHSBALLEY (approved):

PR #1014, opened 2022-11-21
Lbhsballey geolive add schemas

LBHSBALLEY:
Add boundaries to the unrestricted area from 2 more Geolive schemas: Education and Housing. 
Also, change Parking boundaries ingestion method so it works as for boundaries from other schemas.


LBHSBALLEY:
This PR was deleting a connection that cannot be deleted. Cancelling the PR.

PR #1066, opened 2022-12-16
add .pre-commit-config.yaml

timburke-hackit:
adding the file created for gitguardian commit hooks


LBHTKarki (approved):
We can do this until we complete the common setup for the hook

PR #1070, opened 2023-01-04
Dpp 196 watermarks table

timburke-hackit:


timburke-hackit:
need to pair the kms actions back a bit - I was running into problems with an error:
`The new key policy will not allow you to update the key policy in the future.`

terraform/modules/dynamodb/10-aws-dynamodb-table.tf
```diff
@@ -0,0 +1,179 @@
...
+    create = lookup(var.timeouts, "create", null)
+    delete = lookup(var.timeouts, "delete", null)
+    update = lookup(var.timeouts, "update", null)
+  }
+}
+
+resource "aws_dynamodb_table" "autoscaled" {
+  count = var.create_table && var.autoscaling_enabled ? 1 : 0
+
+  name             = var.name
```
LBHTKarki, line 90: prefix here would help prevent clashes

LBHTKarki (approved):

PR #1071, opened 2023-01-04
Create parking_match_pcn_permit_vrm_with_address_match_llpg_nlpg_post…

daclarkelbh:
…codes.py

Created script for matching pcn permit vrm and postcodes to NLPG and LLPG.


LBHSBALLEY (changes requested):
Proposing a few changes to make the job much quicker: 
- Instead of loading 30 days of LLPG/NLPG and 7 days of Parking data, omlly load the latest partition by changing the pushdown predicate

- after we've done that, we can remove the 'where import_date = select max...' in all our SQL queries, because we only have the latest date loaded.

- In the SQL query: instead of selecting 3 times the same LLPG/NLPG data and calling  them _reg _curr and _per, selecting it only once and joining it 3 times with the parking data, using different aliases and different columns as a pivot.

I've cloned the job and written a variation, please check it out and see if the result is correct.
(https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/parking_match_pcn_permit_vrm_with_address_match_llpg_nlpg_postcodes_v2_San_script/script)

barnesm707 (approved):
Changes seem to be very useful

LBHSBALLEY (approved):

PR #1093, opened 2023-01-12
DPP 332 Glue Failure Notifications in GChat

timburke-hackit:
## Link to JIRA ticket

[DPP-332](https://hackney.atlassian.net/browse/DPP-332)

### *What is the problem we're trying to solve*

Alerting for Glue job failures currently generates an email notification, the team feels these emails are not the best way of communicating these alerts and a Google Chat group would facilitate better visibility and collaboration addressing failures.  

### *What changes have we introduced*

This branch adds a lambda that is targeted by the existing Cloudwatch event for Glue failures and will publish details of the failure from that event to a Google Chat group. 

Updates include:

- a new lambda for handling the Cloudwatch event and sending notifications to a Google Chat group
- additional target for the Cloudwatch event to trigger the Lambda
- creation of a new IAM role to run the Lambda and permissions for that role to read the webhook address from Secrets Manager

#### _Checklist_

- [x] Added tests to cover all new production code
- [x] Checked all code for possible refactoring
- [x] Code pipeline builds correctly

### *Follow up actions after merging PR*

- Give users permissions to Google space, so they can see the alarms

[DPP-332]: https://hackney.atlassian.net/browse/DPP-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/terraform/core/28-glue-error-notifications.tf
+++ b/terraform/core/28-glue-error-notifications.tf
@@ -1,6 +1,9 @@
 locals {
-  lambda_timeout = 900
+  lambda_timeout                                   = 900
+  glue_job_failure_gchat_notifications_secret_name = "${local.short_identifier_prefix}gchat-glue-failure-webhook-admin"
 }
+
+data "aws_caller_identity" "current" {}
 
 data "archive_file" "glue_job_failure_notification_lambda" {
   type        = "zip"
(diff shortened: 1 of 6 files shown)


terraform/core/28-glue-error-notifications.tf
```diff
@@ -211,3 +214,116 @@ resource "aws_sns_topic_subscription" "admin_failure_notifications" {
...
+    ]
+  }
+
+  statement {
+    actions = [
+      "kms:GenerateDataKey*",
+      "kms:Decrypt"
+    ]
+    effect    = "Allow"
+    resources = ["*"]
```
LBHTKarki, line 274: could this be limited to just the secrets manager key?

PR #1129, opened 2023-01-30
Delete 17-mtfh-tables-ingestion.tf

daroclark:
Deleting this as the tenureinformation table is now in the overall load.


daroclark:
Too drastic

PR #1165, opened 2023-02-10
RBAC configuration for redshift

LBHTKarki:
This update adds the initial roles configuration for Redshift.

All permissions are based on the current departmental users permissions. Role names match the current user names indicating that they are meant to be department specific.

Configuration also includes roles inheritance and currently we have three levels of roles: public, departmental and data and insight. This ensures invidual permissions don't have to be repeated in multiple places.

The initial deployment will create the roles and grant appropriate permissions for them and the role grants will be added in the following PR (code is in place, just commented out).


LBHTKarki:
closing to investigate test failures in the pipeline

PR #1185, opened 2023-02-20
DPP-97 Revert back to lb_cookie

LBHTKarki:
app_cookie requires additional configuration missing from previous change, so this update will revert the change for now.


timburke-hackit (approved):

LBHTKarki:
This has been fixed forward in other PR, so this one is redundant.

PR #1191, opened 2023-02-22
Changed Lambda Ingestion to output JSON properly

HuuHieuDo:
- Lambda limit set to 15 days now
- JSON outputs correctly instead of Python

### DIFF
--- a/lambdas/vonage_api_ingestion/main.py
+++ b/lambdas/vonage_api_ingestion/main.py
@@ -3,8 +3,7 @@ sys.path.append('./lib/')
 
 import requests
 import json
-from datetime import date
-from datetime import datetime
+from datetime import date, datetime
 from dateutil.relativedelta import *
 import time
 import logging
--- a/lambdas/vonage_api_ingestion/main.py
+++ b/lambdas/vonage_api_ingestion/main.py
@@ -1,25 +1,18 @@
 import sys
-
 sys.path.append('./lib/')
 
 import requests
 import json
-
 from datetime import date
 from datetime import datetime
 from dateutil.relativedelta import *
-
 import time
 import logging
-
 import boto3
+
 from dotenv import load_dotenv
 from os import getenv
-
 import re
-
-logger = logging.getLogger()
-logger.setLevel(logging.INFO)
 
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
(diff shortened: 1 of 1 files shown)


lambdas/vonage_api_ingestion/main.py
```diff
@@ -1,29 +1,22 @@
 import sys
-
 sys.path.append('./lib/')
 
 import requests
 import json
-
 from datetime import date
 from datetime import datetime
```
timburke-hackit, line 7: consider combining these:

```suggestion
from datetime import date, datetime
```

lambdas/vonage_api_ingestion/main.py
```diff
@@ -147,26 +134,24 @@ def create_list_of_call_dates(start_date, end_date):
...
     while date_counter < end_call_date:
-        # print(f'Date Added {str(date_counter)}')
         between_dates.append(str(date_counter))
         date_counter = date_counter + relativedelta(days=+1)
     return between_dates
-    # read the dates as actual dates
 
 
 def single_digit_to_zero_prefixed_string(value):
     return str(value) if value > 9 else '0' + str(value)
```
timburke-hackit, line 149: you don't really need to do this you can use the built in zfill() or formatting with f strings
[https://docs.python.org/3/library/stdtypes.html#str.zfill](https://docs.python.org/3/library/stdtypes.html#str.zfill)

value.zfill(2)

lambdas/vonage_api_ingestion/main.py
```diff
@@ -234,122 +220,94 @@ def list_subfolders_in_directory(s3_client,bucket,directory):
...
 
-      list_of_raw_dates = sorted(list_of_raw_dates, key=lambda date: datetime.strptime(date, "%Y-%m-%d"), reverse=True)
+        largest_value = list_of_raw_dates[0]
+        return largest_value
 
-      # print(list_of_raw_dates)
-      largest_value = list_of_raw_dates[0]
-      return largest_value
 
 # Takes a list of "importyear=X" values and finds the largest one
```
timburke-hackit, line 272: consider making this a docstring for the function rather than a comment 
https://peps.python.org/pep-0257/#what-is-a-docstring

timburke-hackit (approved):

PR #1197, opened 2023-03-01
Made own Landing to Raw Parquet Script

HuuHieuDo:


scripts/jobs/customer_services/vonage_landing_to_raw.py
```diff
@@ -1,60 +1,218 @@
...
-    clean_column_names, get_max_date_partition_value_from_glue_catalogue
+from scripts.helpers.helpers import PARTITION_KEYS
+
+# Base Loads
 
 s3_client = boto3.client('s3')
+
+landing_zone_bucket = "dataplatform-stg-landing-zone"
+raw_zone_bucket = "dataplatform-stg-raw-zone"
+prefix = "customer-services/manual/vonage"
```
timburke-hackit, line 23: these should be env variables - they'll be different in prod, stg and dev

scripts/jobs/customer_services/vonage_landing_to_raw.py
```diff
@@ -1,60 +1,218 @@
...
+                                  repl="")
+            list_of_raw_dates.append(importstring)
+
+        list_of_raw_dates = sorted(list_of_raw_dates, key=int, reverse=True)
+
+        largest_value = list_of_raw_dates[0]
+        return largest_value
+
+
+def get_latest_raw_zone_partition_date(s3, bucket, prefix):
```
timburke-hackit, line 94: should this argument be s3_client rather than s3?

scripts/jobs/customer_services/vonage_landing_to_raw.py
HuuHieuDo: Yeah thats true ill update these now to accept env variables from glue

timburke-hackit (approved):

PR #1203, opened 2023-03-08
Lbh vonage glue one time

HuuHieuDo:


scripts/jobs/customer_services/vonage_one_time_ingestion.py
```diff
@@ -45,10 +45,7 @@ def get_auth_token(client_id, client_secret, scope):
     print(f"Auth Token status Code = {r.status_code}")
     auth_token = json.loads(r.text)
 
-    if(r.status_code == 200):
-        return auth_token
-    else:
-        r.raise_for_status()
+    return auth_token
```
timburke-hackit, line 48: is there a reason you're removing this status code check?

HuuHieuDo: Its gone in my branch, I think we should put it back in though.

timburke-hackit (approved):

PR #1208, opened 2023-03-10
Rewrite of Vonage Landing to Raw, and moving to Pre-prod

HuuHieuDo:


scripts/jobs/customer_services/vonage_landing_to_raw.py
```diff
@@ -1,213 +1,158 @@
...
+    print('Dynamic Frame Loaded')
 
-        latest_date = f'{latest_year}{latest_month}{latest_day}'
-        print(f'The Raw Import Date is {latest_date}')
+    print('Converting the DynamicFrame back to SparkDF')
+    dyf = dyf.toDF()
+    dyf = dyf.withColumn("import_date", lit(key_import_date))
+    dyf = dyf.withColumn("import_day", lit(import_day))
+    dyf = dyf.withColumn("import_month", lit(import_month))
+    dyf = dyf.withColumn("import_year", lit(import_year))
```
timburke-hackit, line 95: you should chain these rather than evaluating after each one

dyf.withCoulumn(...).withCoulumn(...).withCoulumn(...)

I'd also consider assigning it to something that makes it clear you're no longer using a dynamic frame

scripts/jobs/customer_services/vonage_landing_to_raw.py
```diff
@@ -1,213 +1,158 @@
...
-        largest_value = list_of_raw_dates[0]
-        return largest_value
-
-
-def get_latest_raw_zone_partition_date(s3_client, bucket, prefix):
-    # Get Year
-    folder_path = f'{prefix}/'
-
-    year_subfolders = list_subfolders_in_directory(s3_client, bucket, folder_path)
+def read_and_write_partition(landing_zone_bucket, partition_path, raw_zone_bucket, raw_zone_prefix):
```
timburke-hackit, line 67: This would be hard to test. Maybe read the frame, add the new columns, write the frame rather than wrapping it all in one?

HuuHieuDo: I could seperate the functions so that its the one function does one thing rule. I was super wary of memory though. Ill change this

scripts/jobs/customer_services/vonage_landing_to_raw.py
HuuHieuDo: I think doing something like dyf = etc
then sparkDF = asdfljksdf(dyf) means I have both the dyf and the sparkDF in memory?

PR #1230, opened 2023-03-28
create new lambda module

timburke-hackit:
Create a new module to deploy lambda functions with a cron scheduled Cloudwatch event trigger.


terraform/modules/aws-lambda-with-cron-schedule/20-lambda.tf
```diff
@@ -0,0 +1,40 @@
...
+  source = data.archive_file.lambda.output_path
+  etag   = filemd5(data.archive_file.lambda.output_path)
+  acl    = "private"
+}
+
+resource "aws_lambda_function" "lambda" {
+  function_name     = lower("${var.identifier_prefix}${var.function_name}")
+  role              = aws_iam_role.lambda.arn
+  handler           = "main.lambda_handler"
+  runtime           = "python3.8"
```
LBHTKarki, line 23: For flexibility would it be better to set this from input variable and support 3.9 as well if someone wants to use that? No point supporting 3.7 anymore I don't think.

terraform/modules/aws-lambda-with-cron-schedule/20-lambda.tf
```diff
@@ -0,0 +1,40 @@
+locals {
+  lambda_name_underscore = replace(lower(var.function_name), "/[^a-zA-Z0-9]+/", "_")
+}
+
+data "archive_file" "lambda" {
```
LBHTKarki, line 5: Using libraries outside Lambda runtime won't be supported, but this is a good starting point for typical setups.

PR #1243, opened 2023-04-06
Add support for additional bucket policy statements

LBHTKarki:
Update the S3 bucket module so that additional policy statements can be passed to the bucket policy document.

This enables more specific policies to be used for each bucket that are using this module.


timburke-hackit (approved):

LBHTKarki:
This is missing key share handling without which the bucket cannot really be shared. I'm closing this one and will create a new PR with key handling in place.

PR #1251, opened 2023-04-12
Active persons records terraform

annajgibson:
Purpose:
To schedule, trigger and configure the deployment of the active_person_records.py script on a daily basis. Data is read and reshaped into a single refined dataset.


timburke-hackit (approved):
looks fine to me

PR #1275, opened 2023-04-21
Give data and insight read access to landing zone

LBHTKarki:
Update D&I glue role with read access to landing zone


LBHTKarki:
This alone is not going to be enough, so will create a new PR for better access coverage

PR #1293, opened 2023-05-03
DPP-463 Remove obsolete tasks

LBHTKarki:
Remove obsolete tasks from prod. Both data sync and data clean up tasks have been deployed using new setups, so these tasks are no longer needed. We can tidy up the resources (old role and references to it in workflows etc.) further once these have been removed.


LBHTKarki:
closing this as it needs more tidying up first

PR #1326, opened 2023-05-16
update raw zone resources

timburke-hackit:
Previous hard coded values were preventing deployment of core to dev as these resources would then be invalid. 

Auto-formatter has made this look like much larger diff than it is - only functional changes are L198-199.


timburke-hackit:
Auto-formatter has made this look like much larger diff than it is - only functional changes are L198-199.

PR #1364, opened 2023-06-09
Add flexible execution variables to AWS Glue Job Module

timburke-hackit:
Allows flexible execution for AWS Glue Jobs. 

This is a feature of Glue version 3+. The default behavior remains as is for the Production environment for all versions of Glue - standard execution. In Dev and Pre-Prod environments execution will now default to flexible, except in the case of legacy Glue version 2 jobs which aren't able to make use of the flexible execution, these will remain standard.


LBH-wgreeff (approved):
LGTM

PR #1365, opened 2023-06-13
Specify Glue Job Worker Type

timburke-hackit:
Adds worker type arguments that were taking null values and preventing flex exectuition in pre-prod. 

Removes an unnecessary test job from pre-prod.


LBH-wgreeff (approved):
LGTM

PR #1371, opened 2023-06-16
set legacy cte precedence policy 

timburke-hackit:
Setting conf job parameters to account for changes in Spark version. Legacy argument allows for behavior as in Spark 2.4 (Glue 2.0).

From docs

In Spark 3.0, spark.sql.legacy.ctePrecedencePolicy is introduced to control the behavior for name conflicting in the nested WITH clause. By default value EXCEPTION, Spark throws an AnalysisException, it forces users to choose the specific substitution order they wanted. If set to CORRECTED (which is recommended), inner CTE definitions take precedence over outer definitions. For example, set the config to false, WITH t AS (SELECT 1), t2 AS (WITH t AS (SELECT 2) SELECT * FROM t) SELECT * FROM t2 returns 2, while setting it to LEGACY, the result is 1 which is the behavior in version 2.4 and below.

https://spark.apache.org/docs/latest/sql-migration-guide.html#query-engine

### DIFF
--- a/terraform/etl/38-aws-glue-job-parking.tf
+++ b/terraform/etl/38-aws-glue-job-parking.tf
@@ -886,7 +886,7 @@ module "parking_foi_pcn_gds_daily_summary" {
   job_parameters = {
     "--job-bookmark-option" = "job-bookmark-disable"
     "--environment"         = var.environment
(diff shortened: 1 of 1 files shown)


LBH-wgreeff (approved):
LGTM

PR #1381, opened 2023-06-29
Created a job that standardises a raw electoral register input into refined

annajgibson:
Data is transformed into format/schema that matches active_person_records - electoral_register_data_to_refined.py

* also added __init__.py to the data_and_insight folder
* added a new terraform module to set up job and crawler.

### DIFF
--- a/terraform/etl/46-aws-glue-job-person-record-matching.tf
+++ b/terraform/etl/46-aws-glue-job-person-record-matching.tf
@@ -1,12 +1,8 @@
-locals {
-  electoral_register_environment_count = local.is_live_environment && !local.is_production_environment ? 1 : 0
-}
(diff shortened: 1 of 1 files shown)


terraform/etl/46-aws-glue-job-person-record-matching.tf
```diff
@@ -0,0 +1,42 @@
+locals {
```
timburke-hackit, line 1: I'd do away with this and just do the count directly in the module block

timburke-hackit: yup - will let the plans run but lgtm

timburke-hackit (approved):

PR #1382, opened 2023-06-29
Fixes problems with functions in person_matching_module.py which were wrongly configured for ereg data.

annajgibson:
Amends terraform 46-aws-glue-job-person-record-matching.tf to provide catalog database and table arguments instead of S3 path so that the execution context class can be used. Also switches to Glue 3 to avoid abydos dependency and also increases no. workers to 10. Amends electoral_register_data_to_refined.py to use new data arguments,

### DIFF
--- a/lambdas/sftp_to_s3/index.js
+++ b/lambdas/sftp_to_s3/index.js
@@ -7,7 +7,7 @@ const AWS_REGION = "eu-west-2";
 
 const s3Bucket = process.env.S3_BUCKET;
 const s3TargetFolder = process.env.S3_TARGET_FOLDER;
-let sftpFilePath = process.env.SFTP_TARGET_FILE_PATH; 
+let sftpFilePath = process.env.SFTP_TARGET_FILE_PATH;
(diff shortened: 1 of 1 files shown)


terraform/etl/46-aws-glue-job-person-record-matching.tf
```diff
@@ -8,8 +8,8 @@ module "electoral_register_refined" {
   glue_scripts_bucket_id         = module.glue_scripts_data_source.bucket_id
   glue_temp_bucket_id            = module.glue_temp_storage_data_source.bucket_id
   glue_job_worker_type           = "G.1X"
-  number_of_workers_for_glue_job = 2
-  glue_version                   = "4.0"
+  number_of_workers_for_glue_job = 10
```
timburke-hackit, line 11: Just out of curiosity - did you run into problems with fewer?

timburke-hackit (approved):
lgtm

daroclark (approved):

PR #1389, opened 2023-07-05
Dpp-518 Move Academy Rev and Bens to Glue 4.0

timburke-hackit:
Tested in Pre. Ingestion jobs from the database will work as is with a Spark upgrade, copy to raw jobs require legacy time parsing.


LBH-wgreeff (approved):
LGTM

PR #1390, opened 2023-07-05
Dpp-519-specify-glue-version

timburke-hackit:
Moves jobs to version 4 in pre-production for testing and specifies version 2 in production to allow for changing the default module value.


LBH-wgreeff (approved):
LGTM

PR #1399, opened 2023-07-13
Enable Glue Job Metrics

timburke-hackit:
From the console this is on by default. Setting this argument enables Glue Job Metrics to help identify behaviour in the driver and executors.


LBH-wgreeff (approved):
LGTM

PR #1408, opened 2023-08-03
Add pre-commit-config.yaml

timburke-hackit:
Removes pre-commit-config.yaml from .gitignore so that this can be included in the repo. Adds a pre-commit-config that runs GitGuardian / ggshield checks.


LBH-wgreeff (approved):
LGTM

PR #1409, opened 2023-08-08
Set Autoscaling Value for RDS Instance

timburke-hackit:
This sets the maximum size the instance can scale to. 

The Liberator backdated ingestion was failing because there was insufficient storage space, this allows for the instance to expand and complete the task.


LBH-wgreeff (approved):
LGTM

PR #1414, opened 2023-08-10
Update parking_correspondence_performance_qa_with_totals_gds.py

daclarkelbh:
Corrected error on 3 lines as had extra prefix for field name.  Tested in pre-prod again and works


timburke-hackit (approved):
do you mind if I change the commented lines too?

like this
--Nohaad Al-othmani when qa_qa_doc_created_by like '' then ''
to
--Nohaad Al-othmani when qa_doc_created_by like '' then ''

PR #1419, opened 2023-08-21
add policy statements for mtfh

timburke-hackit:
These changes allow the assumed reporting role in the housing production account to:
- put objects in the landing zone of prod and stg under the path /mtfh/*
- use the KMS key for the landing zone buckets in prod and stg to generate a data key and decrypt it

This is so that the assumed role can save a point in time recovery of the MTFH tables in to the landing zone for these two accounts.


LBH-wgreeff (approved):
LGTM

PR #1425, opened 2023-08-30
Fixes to staff sickness forecasting scripts.

annajgibson:
- Added statsmodels==0.14.0 to additional python module to avoid _collections_ import issue.
- Added required date format parameter to reshape_time_series_data function call.
- Converted column names to lowercase to reflect datacatalog format.


timburke-hackit (approved):
lgtm

PR #1427, opened 2023-09-01
Fixes to staff sickness and time series helpers

annajgibson:
- ensuring variables are set to int where needed
- increasing number of workers
- adding a target bucket job parameter for image files, and to associated functions
- adding a method for writing images to s3
- converting pandas df to a spark df before writing to S3

### DIFF
--- a/scripts/helpers/time_series_helpers.py
+++ b/scripts/helpers/time_series_helpers.py
@@ -2,27 +2,26 @@
 Functions to support development of time series analytics work.
 """
 import datetime
-import io
-from typing import Tuple
 
+import io
 import boto3
 import numpy as np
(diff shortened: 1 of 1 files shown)


scripts/helpers/time_series_helpers.py
```diff
@@ -2,8 +2,10 @@
 Functions to support development of time series analytics work.
 """
 import datetime
-from typing import Tuple, Union
+import io
+from typing import Tuple
```
timburke-hackit, line 6: I think you don't need this anymore if you use tuple throughout rather than Tuple - sorry, realise this one came from me in the first place!

timburke-hackit (approved):

PR #1428, opened 2023-09-05
Changes to time series helpers and sickness forecasting scripts

annajgibson:
- removed reference to 'Tuple', and replaced with 'tuple'
- corrected file path for PNG images to include department
- amended bucket reference in terraform module
- amended docstrings where parameters removed/changed


timburke-hackit (approved):
lgtm

PR #1435, opened 2023-09-08
Stop prod key sharing

LBHTKarki:
Key share is no longer necessary, so remove the config for it.


timburke-hackit:
@LBH-wgreeff I left this open for when you came back. Think Daro's done with the work that required this so, OK to merge this in?

timburke-hackit (approved):

PR #1436, opened 2023-09-12
increase parking job workers 

timburke-hackit:
Increases the number of Glue workers from 2 to 6. This job runs monthly and has been timing out at 2 hours, increasing the workers has been tested to show this will reduce the run time beneath the threshold. 

Additionally corrects a spelling mistake in the python script name and job title from tends to trends.


LBH-wgreeff (approved):
LGTM

PR #1438, opened 2023-09-13
Added holt winters to helpers

HuuHieuDo:


scripts/helpers/time_series_helpers.py
```diff
@@ -363,11 +368,44 @@ def forecast_ets(dataframe, start_date, end_date, seasonal=None, damped_trend=Fa
...
+
+            Args:
+                Dataframe (Dataframe): Dataframe containing training timeseries dataset. Must be equally spaced with Date in the Index
+                forecast_count (int): Amount of data points to forecast
+                seasonal (String): Trend Component model. Optional. "Add", "mul" or None (default)
+                seasonal_periods (int): The number of periods in a complete seasonal cycle for seasonal (Holt-Winters) models. For example, 4 for quarterly data with an annual cycle or 7 for daily data with a weekly cycle. Required if seasonal is not None.
+                use_boxcox (bool): None (default) or True
+                initialization_method (String):
+            Returns:
+                Forecast Results (Dataframe),
```
timburke-hackit, line 394: I'm all for auto docstrings, but these aren't right / complete

- Dataframe in the inputs should be dataframe
- initialization_method needs a description
- Return should be pdf and needs a description

timburke-hackit (changes requested):

scripts/helpers/time_series_helpers.py
```diff
@@ -363,11 +368,44 @@ def forecast_ets(dataframe, start_date, end_date, seasonal=None, damped_trend=Fa
...
+        trend="add",
+        seasonal=seasonal,
+        use_boxcox=use_boxcox,
+        initialization_method=initialization_method
+    ).fit()
+
+    forecast = model.forecast(forecast_count)
+
+    pdf = pd.DataFrame()
+    pdf["mean"] = forecast
```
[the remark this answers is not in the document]
timburke-hackit: oh, this isn't appended to the input frame...?

sorry, I think I've misread what you're doing here. 

I'm still not sure it's a good thing to do - why not just pass back the forecast and let people decide what to do with it?

I guess I'm just not understanding why this needs to be a function as opposed to a how to guide.

scripts/helpers/time_series_helpers.py
```diff
@@ -345,59 +331,88 @@ def get_start_end_date(dataframe, period, forecast_count):
 
     start_date = date_maker.get(period)[0]
     end_date = date_maker.get(period)[1]
+
     return start_date, end_date
 
 
-def forecast_ets(dataframe, start_date, end_date, trend="add", seasonal=None, damped_trend=False, seasonal_periods=None):
+def forecast_ets(dataframe: pd.DataFrame, start_date: str, end_date: str, seasonal_period: str = "none",
```
timburke-hackit, line 338: and here

timburke-hackit (approved):
looks  ok to me - just unsure about whether you mean to switch between ps and pd

PR #1442, opened 2023-10-02
Stg Reworked RDS Snapshot to s3

timburke-hackit:
This PR deploys a new module to stg for testing that creates:

- cloudwatch event rule that detects the completion of the liberator RDS snapshot export to s3
- event rule target that invokes a lambda to copy the snapshot export to the raw zone

There are also changes to existing modules:
- output of instance arns from the `sql-to-rds-snapshot` module, these are needed to reference as resources to watch in the new cloudwatch event rule
- update to the `aws-lambda` module so it uses the `s3_key` variable when creating the zipped lambda file resource in s3

This is largely to poc the implementation in stg will invoke the lambda when the ecs task finishes exporting a snapshot from the liberator landing zone bucket. The lambda code in `main.py` is work in progress, I'll raise a separate PR for a more complete version. 

There are also a number of variables in `terraform/modules/rds-snapshot-to-s3/02-inputs-optional.tf` these are for functionality that can be copied over from the existing module `terraform/modules/db-snapshot-to-s3` assuming this implementation works as planned and can replace this module in prod.

### DIFF
--- a/lambdas/s3-to-s3-export-copier-python/main.py
+++ b/lambdas/s3-to-s3-export-copier-python/main.py
@@ -1,0 +1,171 @@
+
+import json
+import boto3
+
+AWS_REGION = "eu-west-2"
+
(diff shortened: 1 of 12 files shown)


LBH-wgreeff (approved):
LGTM

PR #1446, opened 2023-10-04
dpp 558 lambda snapshot s3 to s3 copier

timburke-hackit:
Python code for a lambda that is triggered once a RDS snapshot export has completed. 

This will move the parquet files from the export location to a target location e.g. raw

There were tests for this, but the pipeline runs python 3.8 and this has dependencies on 3.9+.


LBH-wgreeff (approved):
LGTM

PR #1456, opened 2023-10-16
Dpp 558 liberator ingestion

timburke-hackit:
Replaces much of the existing liberator ingestion process.

A new s3 bucket for storing exported snapshots from RDS. This replaces the existing bucket defined in `modules/db-snapshot-to-s3`.

2 new lambdas, python code for these has already been merged into main. These replace the old nodejs lambdas with behaviour that avoids the recursive SQS lambda invocation that motivated the overall rework.

2 eventbridge rules that:

1. detect the completion of the RDS Snapshot creation which in turn invokes a lambda to start the RDS Snapshot export to s3 
2. detect the completion of the export process and invokes a lambda to move the export from the RDS export bucket to the raw zone


LBH-wgreeff (approved):
LGTM

PR #1467, opened 2023-10-25
Cloudwatch vpc flow logs

timburke-hackit:
same as #1465, with the missing IAM role for the aws_flow_log resource


LBH-wgreeff (approved):
LGTM

PR #1470, opened 2023-10-27
Create addressbasepremium_load_files.py

LBHSBALLEY:
2 jobs and 2 look up files to load a full supply of AddressBase Premium GB into the DP. This is a manual process done every 6 weeks:
- Manually drop a full supply into the raw zone/unrestricted/osaddressbasepremium (about 300 csv files), in a new folder indicating the epoch number
- Run the job addressbasepremium_load_files after setting the path of the the raw and the preprocessed files in the parameters
- Run the job addressbasepremium_create_address_table after setting the path of the preprocessed files in the parameters. The look-up files need to be refreshed every year or so.


LBH-wgreeff (approved):
LGTM

PR #1471, opened 2023-10-30
Mtfh state machine

timburke-hackit:
Deploy MTFH state machine to all environments. 

Tidied up the permission resources.  

Enables an Eventbridge rule to trigger the state machine in production. 

This takes an array of table names from an Eventbridge event rule, passes them in a map to a lambda that triggers an export from the DynamoDB tables in the housing account which is then picked up by a subsequent Glue job and moved to the Raw zone.
image


LBH-wgreeff (approved):
LGTM

PR #1472, opened 2023-10-30
Dpp 203 convert csv ingestion to folder based ingestion rather than individual files

LBH-wgreeff:
Convert CSV ingestion to folder-based ingestion rather than individual files

### DIFF
--- a/lambdas/g_drive_folder_to_s3/main.py
+++ b/lambdas/g_drive_folder_to_s3/main.py
@@ -54,10 +54,7 @@ def authenticate_google_drive(google_credentials):
     Args:
         google_credentials (dict): Google Drive API credentials as a dictionary.
 
-    Returns:
-        Credentials: Google Drive API credentials.
     """
-    credentials = None
     creds = None
     SCOPES = ['https://www.googleapis.com/auth/drive', 'https://www.googleapis.com/auth/spreadsheets']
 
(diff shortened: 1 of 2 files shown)


lambdas/g_drive_folder_to_s3/main.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    Authenticate using Google Drive service account credentials.
+
+    Args:
+        google_credentials (dict): Google Drive API credentials as a dictionary.
+
+    Returns:
+        Credentials: Google Drive API credentials.
+    """
+    credentials = None
```
timburke-hackit, line 60: don't think you need this `crentials` variable because you're using the creds one?

lambdas/g_drive_folder_to_s3/test_main.py
```diff
@@ -0,0 +1,115 @@
...
+
+def test_download_file(google_drive_service):
+    # Mock Google Drive service and download response
+    mock_drive_service = google_drive_service
+
+    # Test download_file function
+    file_data = download_file(mock_drive_service, 'your_file_id', 'application/json')
+    assert file_data == b'File Data'
+
+def test_upload_to_s3(s3_bucket):
```
timburke-hackit, line 85: does `test_upload_to_s3` need an assert?

timburke-hackit (approved):

PR #1473, opened 2023-10-31
set account id for environment in Eventbridge rule

timburke-hackit:
Assigns the account id based on environment to be deployed in to.


LBH-wgreeff (approved):
LGTM

PR #1474, opened 2023-10-31
Liberator fixes

timburke-hackit:
- clarified some of the variable names
- policy updates to allow lambdas to do what they're intended to do
- update lambda environment variables


LBH-wgreeff (approved):
LGTM

PR #1475, opened 2023-10-31
Liberator fixes

timburke-hackit:
Correcting policy attachments, actions and role handling in the lambda module.


LBH-wgreeff (approved):
LGTM

PR #1478, opened 2023-11-02
Liberator permissions

timburke-hackit:
IAM updates:
- passrole for RDS export
- trust relationship for export service 
- additional KMS actions to allow use of the bucket key

Lambda Scripts
- parsing of event string to python dict
- logging of event


LBH-wgreeff (approved):
LGTM

PR #1488, opened 2023-11-07
add count parameters

timburke-hackit:
Add count parameters for items that reference resources that will not be created when deploying to the dev account.


LBH-wgreeff (approved):
LGTM

PR #1491, opened 2023-11-08
add rds bucket

timburke-hackit:
This bucket was originally created by the `liberator_db_snapshot_to_s3` module which was removed in (#1490) causing the deployment to want to remove the bucket.


LBH-wgreeff (approved):
LGTM

PR #1493, opened 2023-11-08
Remove academy glue jobs

timburke-hackit:
Removes the Academy table ingestion Glue jobs and triggers that are no longer necessary as they have been replaced by a single job from `module "academy_glue_job"`

Also includes resource index key that was previously (#1492)


LBH-wgreeff (approved):
LGTM

PR #1503, opened 2023-11-13
Move addresses api export bucket

timburke-hackit:
Moves bucket resource out of the db_snapshot_to_s3 module and creates it as its own resource.


LBH-wgreeff (approved):
LGTM

PR #1507, opened 2023-11-21
Add type conversion for event values to be used for calculation

timburke-hackit:
Small addition to handle cases where the event might pass values as strings by converting them to int.


LBH-wgreeff (approved):
LGTM

PR #1508, opened 2023-11-21
add provider

timburke-hackit:
Adds a missing provider parameter to the SSE resource that was preventing the pipeline from running successfully.


LBH-wgreeff (approved):
LGTM

PR #1509, opened 2023-11-22
remove superseded liberator module

timburke-hackit:
The liberator ingestion requiring these resources has been superseded by another implementation so these are no longer required.


LBH-wgreeff (approved):
LGTM

PR #1510, opened 2023-11-22
S3 bucket module changes

timburke-hackit:
Addresses deprecated methods for creating sse config and versioning of s3 buckets by creating these resources separately for all platform buckets. 

This change makes the terraform compliant with the compliance checks made in the workflow that would be triggered were additional buckets to be created.


LBH-wgreeff (approved):
LGTM

PR #1516, opened 2023-11-24
Add academy landing to raw to state machine

timburke-hackit:
Adding the onward crawler and Glue job to the existing database extraction state machine. This will allow better monitoring of the pipeline as a whole and moves to an event driven rather than schedule workflow.


LBH-wgreeff (approved):
LGTM

PR #1519, opened 2023-11-27
replace deprecated sse and versioning

timburke-hackit:
Replaces deprecated arguments in the aws_s3_bucket resource with separate resources for sse and versioning.


LBH-wgreeff (approved):
LGTM

PR #1520, opened 2023-11-27
Dpp 583 Additional Lambda Alarm Notifications

timburke-hackit:
Adds additional lambda alarm notifications for Lambdas in the Academy, Liberator and MTFH workflows.


LBH-wgreeff (approved):
LGTM

PR #1521, opened 2023-11-27
Dpp 584 academy stfn definition template

timburke-hackit:
Moves the state machine definition to a separate template file and uses variables to reference resources. 

Also fixes some state path references. 

This will also make future adaptations to the state machine definition easier to implement.


LBH-wgreeff (approved):
LGTM

PR #1522, opened 2023-11-28
Dpp 584 Academy State Machine Patch

timburke-hackit:
Minor changes to the Academy State Machine:

- removes an erroneous trailing brace on a variable in the definition file
- sets max concurrency for the copy to raw job to be equal to the number of filter strings
- enable auto-scaling of glue workers
- removes trigger for job as this is now orchestrated by the state machine


LBH-wgreeff (approved):
LGTM

PR #1525, opened 2023-11-28
Dpp 584 academy patch

timburke-hackit:
- handling for crawler state to check if crawler is still running / stopping and catch errors
- remove crawler trigger, this is run as part of the orchestrated workflow so doesn't need to be scheduled
- better separation of benefits / revenues datasets


LBH-wgreeff (approved):
LGTM

PR #1528, opened 2023-11-30
Added 82-academy-pre-production-raw-zone-data.tf as a clone of 83-...-test.tf with "test" removed from code

stevefarrhackneygovuk:
Reinistating the the crawlers for Academy raw zones in Pre-Production only, to catch the data that is copied over from Production.

### DIFF
--- a/terraform/core/82-academy-pre-production-raw-zone-data.tf
+++ b/terraform/core/82-academy-pre-production-raw-zone-data.tf
@@ -29,7 +29,7 @@ resource "aws_glue_crawler" "revenue_raw_zone" {
 
         Partitions = {  
             AddOrUpdateBehavior = "InheritFromTable"  
-         }
+        }
     })
 }
 
@@ -62,6 +62,6 @@ resource "aws_glue_crawler" "bens_housing_needs_raw_zone" {
 
         Partitions = {  
             AddOrUpdateBehavior = "InheritFromTable"  
-         }  
+        }  
     })
 }


terraform/core/82-academy-pre-production-raw-zone-data.tf
```diff
@@ -0,0 +1,55 @@
...
+
+    s3_target {
+        path = "s3://${module.raw_zone.bucket_id}/revenues/"
+    }
+
+    configuration = jsonencode({
+        Version = 1.0
+        Grouping = {
+            TableLevelConfiguration = 3
+        }
```
timburke-hackit, line 26:
```suggestion
    TableGroupingPolicy     = "CombineCompatibleSchemas"
        }
```
This is equivalent to the one table per s3 path configuration and should prevent the crawler from creating multiple versions of a table

terraform/core/82-academy-pre-production-raw-zone-data.tf
```diff
@@ -0,0 +1,55 @@
...
+
+    s3_target {
+        path = "s3://${module.raw_zone.bucket_id}/benefits-housing-needs/"
+    }
+
+     configuration = jsonencode({
+        Version = 1.0
+        Grouping = {
+            TableLevelConfiguration = 3
+        }
```
timburke-hackit, line 53:
```suggestion
    TableGroupingPolicy     = "CombineCompatibleSchemas"
        }
```
This is equivalent to the one table per s3 path configuration and should prevent the crawler from creating multiple versions of a table

terraform/core/82-academy-pre-production-raw-zone-data.tf
```diff
@@ -0,0 +1,55 @@
...
+    })
```
timburke-hackit, line 54:
```suggestion
  Partitions = {
          AddOrUpdateBehavior = "InheritFromTable"
          }
    })
```
This prevents the crawler from inferring different schemas and will take the definition from the catalog

terraform/core/82-academy-pre-production-raw-zone-data.tf
```diff
@@ -0,0 +1,55 @@
...
+    s3_target {
+        path = "s3://${module.raw_zone.bucket_id}/revenues/"
+    }
+
+    configuration = jsonencode({
+        Version = 1.0
+        Grouping = {
+            TableLevelConfiguration = 3
+        }
+    })
```
timburke-hackit, line 27:
```suggestion
  Partitions = {
          AddOrUpdateBehavior = "InheritFromTable"
          }
    })
```
This prevents the crawler from inferring different schemas and will take the definition from the catalog

timburke-hackit (approved):

PR #1530, opened 2023-12-04
add export bucket kms arn

timburke-hackit:
s3-to-s3 copier lambda is currently failing in the production api account with an AccessDenied error. This change will allow the lambda role to decrypt the exported RDS snapshot files.


LBH-wgreeff (approved):
LGTM

PR #1531, opened 2023-12-04
Revert #1528

timburke-hackit:
Reverts #1528  which adds resources which conflict with manually created resources in stg


LBH-wgreeff (approved):
LGTM

PR #1538, opened 2023-12-07
avoid conflicting names

timburke-hackit:
This module was trying to create an already existing IAM role, name changed to avoid this.


LBH-wgreeff (approved):
LGTM

PR #1547, opened 2023-12-20
Dpp 597 liberator notifications

timburke-hackit:
Adds a terraform module for receiving email alerts for files being added to s3, as implemented for the liberator database dump.


LBH-wgreeff (approved):
LGTM

PR #1551, opened 2024-01-15
Dpp 573 parking tables load

timburke-hackit:
Python code to run a procedure that will copy data from a location in s3 to a table in Redshift.

A sql procedure to stage and replace data in a Redshift table from s3. 

Terraform to deploy the lambda code, attach a role to the redshift cluster that can access the data in s3, an IAM role and policies to execute the lambda that can retrieve that database user details from Secrets Manager and execute commands on the Redshift cluster.


LBH-wgreeff (approved):
LGTM

PR #1556, opened 2024-01-17
add read policies for departments to access redshift

timburke-hackit:
Adds read access policies to the department sso groups to allow users to see the clusters in the console and access the query editor. 

This doesn't grant any access to data from the cluster without a valid db user login. Only changes staging for testing.


LBH-wgreeff (approved):
LGTM

PR #1557, opened 2024-01-17
Purpose: To clean addresses for linkage work

tomc-hackney:
Python code that cleans and standardises full address data into a format optimised for record matching


LBH-wgreeff (approved):
LGTM

scripts/jobs/data_and_insight/address_cleaning.py
```diff
@@ -0,0 +1,131 @@
...
+
+        for reg in inputs.clean_trim_address_regex_subs:
+            source_df = source_df.withColumn("clean_trim_address",
+                                             F.regexp_replace(F.col("clean_trim_address"), reg[0], reg[1]))
+
+        # tidy up extra spaces
+        source_df = source_df.withColumn("clean_trim_address",
+                                         F.regexp_replace(F.col("clean_trim_address"), " ?- ?$", ""))
+        source_df = source_df.withColumn("clean_trim_address", F.trim(F.col("clean_trim_address")))
+
```
LBHSBALLEY, line 119: Lines 116-118 would make a nice function

scripts/helpers/address_cleaning_inputs.py
```diff
@@ -0,0 +1,103 @@
+"""Subs list for address cleaning"""
+
```
LBHSBALLEY, line 2: In all the regex I would use "\s" instead of " " because it also covers white tabs

LBHSBALLEY (commented):
Looks good but could use more regex to make the script shorter

PR #1558, opened 2024-01-17
Dpp 594 department redshift console permissions

timburke-hackit:
More read permissions for departments accessing redshift through the console - sorry, this might be a bit whack a mole.


LBH-wgreeff (approved):
LGTM

PR #1559, opened 2024-01-17
Create lambda layer module

LBH-wgreeff:
This is a new module that can be used to create layers from Lambdas


terraform/modules/aws-lambda-layers/02-inputs-optional.tf
timburke-hackit, line 1: you don't need all of these in this module, right?

terraform/core/44-lambda-layers.tf
timburke-hackit, line 1: this should probably be in etl or they'll be difficult to reference from that state where the lambdas should probably be.

terraform/modules/aws-lambda-layers/10-lambda-layer.tf
```diff
@@ -0,0 +1,25 @@
...
+  lambda_name_underscore = replace(var.lambda_name, "-", "_")
+  command                = "make all"
+  environment_map        = var.environment_variables == null ? [] : [var.environment_variables]
+}
+
+resource "null_resource" "run_install_requirements" {
+  count = var.install_requirements ? 1 : 0
+  triggers = {
+    dir_sha1 = sha1(join("", [for f in fileset(path.module, "../../../lambdas/${local.lambda_name_underscore}/*") : filesha1("${path.module}/${f}")]))
+  }
```
timburke-hackit, line 11: I don't think you need any of this either?

PR #1563, opened 2024-01-18
update number of splits calculation

timburke-hackit:
Amending calculation for the number of splits to be used when reading from DynamoDB. 

This is now done dynamically based on the number and type of executors assigned to the Glue job.


LBH-wgreeff (approved):
LGTM

PR #1578, opened 2024-02-02
remove the typo "and"

TianChen17:
Glue failure detected for job: parking_cycle_hangar_met_fail_monthly_format.

This Glue Job fails; since there has been no recent change to the Python script, I suspect there might be a minor issue with the SQL. There is a typo in "and".

I attempted to run this SQL in Athena but am not entirely sure how Mike checked it before merging it into the Main branch.


barnesm707 (approved):
No idea how that got there, but that will be the problem

PR #1583, opened 2024-02-06
Revert "Dpp 533 testing export connector for mtfh dynamodb tables"

timburke-hackit:
The assumed role in the dynamodb account lacks permissions to write to the temporary Glue storage in our account.


TianChen17 (approved):
ok

PR #1586, opened 2024-02-07
add layers variable

timburke-hackit:
Allows the attachment of layers to deployed AWS Lambda functions using a _layers_ variable which takes a list of arns.


LBH-wgreeff (approved):
LGTM

PR #1590, opened 2024-02-08
Update 08-google-sheets-imports.tf

barnesm707:
need to ingest cycle hangar installation details also have put correct worksheet name into earlier entry (for suspension data)


LBH-wgreeff (approved):
LGTM

PR #1593, opened 2024-02-13
add comments and measure the time cost - on parking pcn daily

TianChen17:
I have a ticket [https://hackney.atlassian.net/browse/DPP-618] to understand the bottleneck of our current parking glue jobs, which use a significant amount of resources (10 workers) and also take a considerable amount of time.

This change measures the time cost of each snippet. I have first tested it in pre-production, and all looks good. [https://eu-west-2.console.aws.amazon.com/gluestudio/home?region=eu-west-2#/editor/job/stg%20parking_pcn_daily_print_monitoring--check-time-cost/runs]


LBH-wgreeff (approved):
LGTM

PR #1581, opened 2024-02-05
Dpp 533 testing export connector for mtfh dynamodb tables

timburke-hackit:
Changes the dynamodb connection options to use the export connector rather than the ETL connector that was previously used. 

This method does not require the splits option so parameters and calculation associated with this have been removed. 

Auto-scaling of workers has been re-enabled to allow for inactive workers to be removed while the export action takes place.

PR #1603, opened 2024-02-20
Dpp 533 testing export connector mtfh

timburke-hackit:
Changes the Glue job to make use of the DynamoDB export connector rather than the ETL connector so that the task no longer needs to scan the DynamoDB tables. 

update of #1581 - now includes necessary bucket target and connection details to allow the export role in the housing account to do the cross account export of the dynamodb tables.


LBH-wgreeff (approved):
LGTM

PR #1606, opened 2024-02-23
only load the needed data and write to S3 using single parquet

TianChen17:
this is a reply to ticket blow
https://hackney.atlassian.net/browse/DPP-649
the code added:
1) add pushdown_predicate
2) merge very small parquet files to 1 (have not defined a function to achieve it, since it's a straightforward way to add `coalescedDF = ApplyMapping_node2.toDF().coalesce(1)` for most of existing ETL scripts).


timburke-hackit (approved):
Looks good to me Tian, 👍.

Can we make sure #1605 is merged before deploying to prod? Thanks

PR #1607, opened 2024-02-26
Fix glue alerts

timburke-hackit:
Gets the job attempt from the Glue API as this isn't present in the cloudwatch event - not sure how I got this confused the first time, sorry!


LBH-wgreeff (approved):
LGTM

PR #1608, opened 2024-02-26
Added in lambda layer for urllib3

annajgibson:
Added in arn:aws:lambda:eu-west-2:120038763019:layer:urllib3-1-26-18-layer:1 to ensure correct urllib3 version used.


LBH-wgreeff (approved):
LGTM

PR #1609, opened 2024-02-26
Dpp 650 redshift role association

timburke-hackit:
Creating an additional _aws_redshift_cluster_iam_roles_ resource is causing the role created by the redshift module to be disassociated with the cluster in stg.

Adds a variable to allow additional IAM roles to be associated with the cluster in the Redshift module.

Removes the _aws_redshift_cluster_iam_roles.parking_redshift_copier_ resource.


LBH-wgreeff (approved):
LGTM

PR #1610, opened 2024-02-27
replace instances of hangar_corral_number with asset_no (DPP-652)

timburke-hackit:
The source data for these jobs has changed with the column _hangar_corral_number_ being renamed _asset_no_ causing the jobs to fail. 

This replaces all isntances of _hangar_corral_number_ with _asset_no_.


LBH-wgreeff (approved):
LGTM

PR #1611, opened 2024-02-27
Changes to GovNotify lambda function:

annajgibson:
Terraform:
- added in correct lambda role for function to use
- specified Python runtime version as 3.9 (instead of default 3.8)
- amended the s3 bucket and output folder strings

Python main.py:
- moved dict and list into main script as could not be found.

Helpers.py
- deleted as no longer needed as a separate file.


LBH-wgreeff (approved):
LGTM

PR #1613, opened 2024-02-28
Dpp 660 extend lambda timeout

timburke-hackit:
Extends the timeout for the Glue failure alerts lambda from 3s to 10s.

Additionally adds the ability to set the timeout in the terraform module.

This is to address a bug where the lambda would successfully send a message but timeout before completing then retry resulting in duplicate messages. The runtime is generally around 2.5s so setting this to 10s should provide ample time to complete.


timburke-hackit:
@LBH-wgreeff dunno why the linter took a fit earlier, but seems happy now

LBH-wgreeff (approved):
LGTM

PR #1620, opened 2024-03-01
add the role and runtime to the lambda function

LBHMVillalobos:
Terraform file - 49-lambda-streetscene-ingestion
   - runtime added
   - role added

Note: There is something wrong with the role and the permissions (unable to retrieve the Secret permission).  Wayne will look at this issue.


timburke-hackit (approved):
one small suggestion but otherwise looks OK to me

PR #1623, opened 2024-03-03
DPP-662 add pushdown predicates

timburke-hackit:
Adds push_down_predicate argument to 23 Parking Glue jobs to limit the data read when creating DynamicFrames.

Extends the buffer from 1 day to 7 in the script `scripts/jobs/parking/parking_pcn_daily_print_monitoring.py`.


LBH-wgreeff (approved):
LGTM

timburke-hackit:
@TianChen17 yeah, I actually thought the same thing and will go through and make it the latest partition rather than just the previous day

timburke-hackit:
@TianChen17 actually maybe there's an edge case where `push_down_predicate=create_pushdown_predicate("import_date", 5),` is safer?

PR #1630, opened 2024-03-05
remove push_down_predicate arguments

timburke-hackit:
Partially reverses #1623  which added `push_down_predicate` arguments to the creation of DynamicFrames. Some of the tables have been partitioned in a way that doesn't include the `import_date` partition and others do not have a new partition created daily and have sparse partition structure. 

These `push_down_predicate` arguments have been:
- commented out for tables that need repartitioning with the intention of reintroducing them once this work has been done. This makes it easier to find the relevant tables and reintroduce them than were they to be removed entirely.
- removed for tables which don't have daily partitions, these should not be reintroduced and are small tables to begin with.

This addresses the 8 failing jobs 05/03/2024


timburke-hackit:
@TianChen17 would you mind reviewing this please so I can get it deployed before tomorrow's runs. Thanks!

PR #1634, opened 2024-03-06
dpp-664 remove push down predicate

timburke-hackit:
Follow up to #1630. Removes a pushdown predicate for a table that doesn't have daily partitions.


LBH-wgreeff (approved):
LGTM

PR #1637, opened 2024-03-07
add redshift command function into helpers

TianChen17:

What I have done is:
1.	Define a get_secret_dict by tweaking the existing get_secret function, which will return credentials in dict instead of a big string.
2.	Define a rs_command function, which allows us to run all the Redshift commands (such as create tables, etc) and return results in a Python script.

In order to connect with Redshift, we need to use psycopg2 by installing psycopg2-binary, which is added to both requirements.build.txt and requirements.test.txt files.

### DIFF
--- a/scripts/requirements.test.txt
+++ b/scripts/requirements.test.txt
@@ -2,6 +2,6 @@ pytest-watch
 freezegun
 pytest-mock
 pydeequ
-urllib3>=1.25.4,<1.27
-# add redshift connector here to pass CI / Test Python Jobs & Lambda Functions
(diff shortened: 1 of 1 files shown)


LBH-wgreeff (commented):
As I mentioned in my message to you the binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

LBH-wgreeff (approved):
I've approved this but please make sure that the binary package is only used for development and testing but in production, we need to use the package built from sources.

LBH-wgreeff (commented):
Great, let's see how that goes. I see the project is also well-maintained.

PR #1638, opened 2024-03-07
create a user for database migration

TianChen17:

This role is created to allow a user to use an access ID and access key to retrieve the Glue data catalogue, and then use it to define the columns of Academy database tables. Afterwards, the role will load the Parquet files in S3 to the Redshift Serverless in a Python script. All access is related to these procedures.

I should limit the Redshift Serverless access a bit, but currently, we are still conducting loads of testing in Redshift. When we officially start using Redshift Serverless in production, we can review this again.


LBH-wgreeff (approved):
LGTM

PR #1639, opened 2024-03-07
load parquet files into redshfit for all academy data

TianChen17:

This script loads all academy data into Redshift Serverless; we have over 200 tables.

This has been tested and is working fine locally. Now, in our staging Redshift Serverless, it already has all the tables. After we schedule it in glue then, it will truncate each time everyday and load the recent parquet files into redshfit.

The 200 tables costs about 50 mintues to complete in my local PC.

Sorry, I'm using print statements instead of logging, and there are two lines of debugging print statements that have not been deleted so far. After I confirm it is working in our Glue, I will make no more changes, and then I will delete them.


scripts/jobs/academy_data/load_all_academy_data_into_redshift.py
```diff
@@ -0,0 +1,143 @@
...
+    process_load_tables(
+        schema='ctax', # Redshift schema
+        catalog='revenues-raw-zone', # Glue catalog database
+        # map the table names in Glue catalog to names in Redshift
+        table_mapping= {
+            'lbhaliverbviews_core_ct': 'ct',
+            'lbhaliverbviews_core_sy': 'sy'
+        },
+        iam_role='arn:aws:iam::120038763019:role/dataplatform-stg-redshift-serverless-role',
+        base_s3_url = "s3://dataplatform-stg-raw-zone/revenues/"
```
timburke-hackit, line 99: it might be a good idea to import these as environment variables - you'll need them to change depending on where your script is deployed anyway.

scripts/jobs/academy_data/load_all_academy_data_into_redshift.py
```diff
@@ -0,0 +1,143 @@
...
+        table_mapping= {
+            'lbhaliverbviews_core_sy': 'sy',
+            'lbhaliverbviews_core_ctaccount': 'ctaccount',
+            'lbhaliverbviews_core_ctnotice': 'ctnotice',
+            'lbhaliverbviews_core_ctoccupation': 'ctoccupation',
+            'lbhaliverbviews_core_ctproperty': 'ctproperty',
+            'lbhaliverbviews_core_cttransaction': 'cttransaction',
+        },
+        iam_role='arn:aws:iam::120038763019:role/dataplatform-stg-redshift-serverless-role',
+        base_s3_url = "s3://dataplatform-stg-raw-zone/revenues/"
```
timburke-hackit, line 129: and these

scripts/jobs/academy_data/load_all_academy_data_into_redshift.py
```diff
@@ -0,0 +1,144 @@
...
+    # for all tables under ctax
+    process_load_tables(
+        schema='ctax', # Redshift schema
+        catalog='revenues-raw-zone', # Glue catalog database
+        # map the table names in Glue catalog to names in Redshift
+        table_mapping= {
+            'lbhaliverbviews_core_ct': 'ct',
+            'lbhaliverbviews_core_sy': 'sy'
+        },
+        iam_role=f'arn:aws:iam::120038763019:role/dataplatform-{environment}-redshift-serverless-role',
```
timburke-hackit, line 99: this won't work in prod or dev, the account ID is for the stg account

if you take the arn from the terraform resource you can set a job parameter that can pass this dynamically. It would also account for if the role name was changed for whatever reason in the resource. 

The same is true for the s3 paths, you could interpolate the bucket name from the resource and append the s3 object prefix to that.

terraform/core/51-load-all-academy-data-into-redshift-serverless.tf
```diff
@@ -0,0 +1,22 @@
+module "load_all_academy_data_into_redshift" {
+  source                         = "../modules/aws-glue-job"
+  is_live_environment            = local.is_live_environment
+  is_production_environment      = local.is_production_environment
+  department                     = module.department_parking_data_source
```
timburke-hackit, line 5: this is the wrong department which will mean the glue role won't have access to the s3 locations and the module will look for the script in /scripts/jobs/department/script_name.py - in this case /scripts/jobs/parking/load_all_academy_data_into_redshift.py  - which doesn't exist.

terraform/core/51-load-all-academy-data-into-redshift-serverless.tf
```diff
@@ -0,0 +1,28 @@
...
+  glue_scripts_bucket_id          = module.glue_scripts.bucket_id
+  spark_ui_output_storage_id      = module.spark_ui_output_storage.bucket_id
+  glue_version                    = "4.0"
+  glue_job_worker_type            = "G.1X"
+  number_of_workers_for_glue_job  = 2
+  glue_job_timeout                = 220
+  schedule                       = "cron(15 7 ? * MON-FRI *)"
+  job_parameters = {
+    "--environment"                      = var.environment
+    "--role_arn"                         = "arn:aws:iam::${var.aws_api_account_id}:role/dataplatform-${var.environment}-redshift-serverless-role",
```
timburke-hackit, line 22:
```suggestion
    "--role_arn"                         = "arn:aws:iam::${data.aws_caller_identity.account_id}:role/dataplatform-${var.environment}-redshift-serverless-role",
```

timburke-hackit (approved):
Looks good to me, good work.

PR #1652, opened 2024-03-18
add glue serverless connection

TianChen17:
This is to attach a subnet (in the same VPC) to this Glue job; otherwise, we cannot connect to the Redshift serverless using redshift_connector, as it shows a timeout error.

I have tried two methods to resolve this:
1. Using the existing JDBC model, which works well (see the [log](https://github.com/LBHackney-IT/Data-Platform/actions/runs/8326026823/job/22780869904)). but it launches many other resources, e.g., aws_glue_catalog_database, Glue crawler, workflow, and related policies.
2. I decided to tailor one specifically for this, which only creates an aws_glue_connection, with a subnet and security group attached.

Thanks for your review.


LBH-wgreeff (approved):
LGTM

PR #1654, opened 2024-03-19
create pass role policy for departments

timburke-hackit:
Users have experienced the following error when attempting to use Glue interactive sessions:
`Following exception encountered while creating session: An error occurred (AccessDeniedException) when calling the CreateSession operation: User: arn:aws:sts::*:assumed-role/dataplatform-stg-glue-data-and-insight/GlueJobRunnerSession is not authorized to perform: iam:PassRole on resource: arn:aws:iam::*:role/dataplatform-stg-glue-data-and-insight because no identity-based policy allows the iam:PassRole action `

These changes allow Glue job runners to pass the department IAM role to the Glue service in order to create sessions and access resources that would be available to a Glue script run in a non-interactive session i.e. when run as a job.

ref: https://docs.aws.amazon.com/glue/latest/dg/glue-is-security.html

### DIFF
--- a/terraform/modules/department/50-aws-iam-policies.tf
+++ b/terraform/modules/department/50-aws-iam-policies.tf
@@ -769,8 +769,7 @@ data "aws_iam_policy_document" "glue_pass_role_to_secrets_manager_for_notebook_u
 }
 
 resource "aws_iam_policy" "glue_pass_role_to_secrets_manager_for_notebook_use" {
-  tags = var.tags
-
+  tags   = var.tags
   name   = lower("${var.identifier_prefix}-${local.department_identifier}-glue-pass-role-to-secrets-manager-for-notebook-use")
   policy = data.aws_iam_policy_document.glue_pass_role_to_secrets_manager_for_notebook_use.json
 }
--- a/terraform/modules/department/50-aws-iam-policies.tf
+++ b/terraform/modules/department/50-aws-iam-policies.tf
@@ -750,8 +750,8 @@ data "aws_iam_policy_document" "redshift_department_read_access" {
 }
 
 
-// Glue Agents can pass their role to Secrets Manager for notebook use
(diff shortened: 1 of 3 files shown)


terraform/modules/department/50-aws-iam-policies.tf
```diff
@@ -758,3 +748,28 @@ data "aws_iam_policy_document" "redshift_department_read_access" {
...
+    condition {
+      test     = "StringEquals"
+      variable = "iam:PassedToService"
+      values   = ["glue.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_policy" "glue_runner_pass_role_to_glue_for_notebook_use" {
+  tags   = var.tags
```
Tian-2017, line 772: we can add "count = var.environment == "prod" ? 0 : 1" here too to avoid create this policy in prod.

Tian-2017 (approved):
Excellent work, Tim; Thanks for your help.

I have one more request:
Could we change the description of the environment variable? It is misleading due to the use of 'Prod' and 'Stg' in uppercase. see [line 3](https://github.com/LBHackney-IT/Data-Platform/blob/8d413fb0e53df106f8b748ec307afaf3718b7b81/terraform/core/01-inputs-required.tf#L3)

terraform/modules/department/50-aws-iam-policies.tf
timburke-hackit: yeah, I'll update the variable description - that's how I landed up with Prod in one of my earlier commits and you're right, I should have updated the description too. 

Creating the policies in prod I'm inclined to leave them. No one should be manually attaching policies in production anyway and it's one less thing to change if we wanted to do a notebook set up for production down the line (but that would obviously need a much more thought through implementation).

Tian-2017: Cheers

PR #1659, opened 2024-03-26
fix the lambda deployment

Tian-2017:
when change the main.py file, it cannot create a new zip file as s3 object, so it will deploy (terraform) the old version zip file to AWS. Now, by adding metadata, make sure it will create new zip file if the source code changes.

Still have an issue using my git in vscode, let me submit the PR using the github UI.


timburke-hackit (approved):
Thanks for this Tian - looks good.

PR #1660, opened 2024-03-26
Change schedule to run 22:00 Monday to Friday, in line with Academy data ingestion schedule.

annajgibson:


Tian-2017 (approved):
looks good

PR #1664, opened 2024-03-27
DPP-661-test-if-new-lambda-fix-works-in-prod

Tian-2017:
A fix has been deployed to lambda module (see [URL ](https://github.com/LBHackney-IT/Data-Platform/pull/1659)), this PR is to change a main.py file by adding an empty line, which will be used to test if the fix works in prod.


LBH-wgreeff (approved):
LGTM

PR #1669, opened 2024-03-28
add in the aws_cloudwatch_event_target block now that the permissions have been granted.

annajgibson:
- add in the aws_cloudwatch_event_target block now that the permissions have been granted.


Tian-2017 (approved):
thanks

PR #1670, opened 2024-04-02
remove ast literal evaluation from parking RDS export lambdas

timburke-hackit:
Address incorrect parsing of the RDS event resulting in the following error. 

`[ERROR] ValueError: malformed node or string:`

### DIFF
--- a/lambdas/export_rds_snapshot_to_s3/main.py
+++ b/lambdas/export_rds_snapshot_to_s3/main.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import ast
 
 import boto3
 from botocore.exceptions import ClientError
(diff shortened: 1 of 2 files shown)


lambdas/export_rds_snapshot_to_s3/main.py
```diff
@@ -13,14 +12,12 @@ def lambda_handler(event, context):
     print("## EVENT")
     print(event)
 
+    snapshot_identifier = event["detail"]["SourceIdentifier"]
+    source_arn = event["detail"]["SourceArn"]
     bucket_name = os.environ["BUCKET_NAME"]
     iam_role_arn = os.environ["IAM_ROLE_ARN"]
     kms_key_id = os.environ["KMS_KEY_ID"]
```
Tian-2017, line 19: Better to move the env variables outside the functions, in order to avoid reading these variables every time the function is invoked

lambdas/export_rds_snapshot_to_s3/main.py
```diff
@@ -1,6 +1,5 @@
 import logging
```
Tian-2017, line 1: Looks behind line 7 needs adding 
`logger.setLevel(logging.INFO)`
otherwise, logger.info in the try and except block won't work, since the default logging level is "warning".  (see this [URL](https://docs.python.org/3/howto/logging.html#:~:text=The%20default%20level%20is%20WARNING%2C%20which%20means%20that%20only%20events%20of%20this%20level%20and%20above%20will%20be%20tracked%2C%20unless%20the%20logging%20package%20is%20configured%20to%20do%20otherwise.)).

Tian-2017 (approved):
Approved. Thanks Tim for fixing it. It only needs a tiny change, then it's ready to merge.

PR #1671, opened 2024-04-02
add pushdown predicates

timburke-hackit:
Add pushdown predicates to limit data load and reduce run time. 
This job is causing the Parking Glue Workflow to run an additional hour.


Tian-2017 (approved):
Thanks

PR #1672, opened 2024-04-02

daclarkelbh:
Updated source file input_file_name                = "cycle_hangar_denormalised_backfill/cycle hangar denormalised - backfill - 20220401 to 20231001 - UTF8.csv"


Tian-2017:
Hey @daclarkelbh , could you adjust your cron time a bit in the future, since AWS Glue jobs has limitation of 100 per account per region. I am concerned  that we may reach the max limit in the future. Thank you.
For example, if this job is not dependent on others, feel free to run it at  `cron(0 20 * * ? *)`

PR #1673, opened 2024-04-03
DPP-673 fix g-drive lambda re-create

Tian-2017:
The logic is `source code change -> tringer  local-exec to install  -> tringer the creatation of archive_file`

1) Added a `null_resource `to ensure that the installation happens only when there are changes to the source code. Then, use 'local-exec' for the required packages.
2) Incorporated `depends_on = [null_resource.run_install_requirements]` to check for the necessity of creation of zip file. If no installation has been done, then the zip file should not be created.

These steps will ensure that a zip file is generated only when there are changes to the source code.


Tian-2017:
@timburke-hackit Hey Tim, could you help to have a review when you have time. Thank you

timburke-hackit (approved):

PR #1675, opened 2024-04-03
change install command from my simple test to prod

Tian-2017:
forgot to change the installation command from my simple test to prod


Tian-2017:
@timburke-hackit Thanks for the prompt reply. 
Good idea to have a general one.

PR #1677, opened 2024-04-04
Update snapshot copier lambda to handle backdated snapshots

timburke-hackit:
Adds handling of backdated snapshots to trigger the appropriate workflow.

Additional changes:
- validation that the snapshot id is in the expected format
- move print statements to log messages
- allow the "glue:UpdateWorkflow" action so that the Lambda can update the `import_date` value in the backdated workflow


timburke-hackit:
somehow moving laptops I landed up with using a different github user (that's also me - I keep meaning to merge them...)

Tian-2017 (approved):
All look good. Thanks

PR #1679, opened 2024-04-04
DPP-673 remove the existing venv

Tian-2017:
Sorry for bothering you so many times. 

There is a new error. I searched online, and it mentioned that venv already exist. I thought the venv would recreate every time in a clean environment when we trigger the CI/CD. It seems this is not the case. (I will spend time on understanding our GitHub runner later)

I will add the line `rm -rf venv/ `to remove the virtual environment if it exists (due to previous failures). If the venv does not exist, this line will not cause an error.

│ Error: local-exec provisioner error
│ 
│   with module.parking_eta_decision_records[0].module.import_file_from_g_drive.null_resource.run_install_requirements[0],
│   on ../modules/g-drive-to-s3/10-lambda.tf line 115, in resource "null_resource" "run_install_requirements":
│  115:   provisioner "local-exec" ***
│ 
│ Error running command 'make install-requirements': exit status 2. Output:
│ python3 -m venv venv
│ Error: [Errno 17] File exists:
│ '/home/runner/work/Data-Platform/Data-Platform/lambdas/g_drive_to_s3/venv'
│ make: *** [Makefile:4: install-requirements] Error 1

### DIFF
--- a/lambdas/g_drive_to_s3/Makefile
+++ b/lambdas/g_drive_to_s3/Makefile
@@ -1,6 +1,7 @@
 .PHONY: install-requirements
 
 install-requirements:
+	rm -rf venv/
 	python3 -m venv venv
 	# the requirements are generated so that the packages
 	# could be downloaded and packaged up for the lambda


timburke-hackit:
Are you maybe running into this after the first Lambda is created? Looking at the other examples they tend to do this as a tidy up after installing the requirements, like this:
https://github.com/LBHackney-IT/Data-Platform/blob/4b0690e69b843e1113e62842f7043aaea24aded5/lambdas/lambda_alarms_handler/Makefile#L10

Tian-2017:
Yes, if you check the entire Makefile, you will notice that it includes `rm -rf venv/ `at the end (after installation completion).

The reason I want to add another  `rm -rf venv/ ` at the beginning is that the logs indicated the virtual environment already exists, which was caused by my previous failed run.

timburke-hackit:
Ah yeah, sorry, I just saw the diffs - makes sense but strange it would be a problem now

PR #1683, opened 2024-04-05
use layer instead of generating in virtual env

Tian-2017:
1.	Add a layer, which include OAuth2Client and Google-API-Python-Client.
2.	I have not deleted the Makefile, Pipfile, etc., since, if we don't use local-exec, they won't be triggered. If all works fine, we can delete them later.
3.	Comment out the import of dotenv, which is used to load the .env file into the environment; we don't need it.

Thanks again, Tim, for your discussion, ideas, and review.

### DIFF
--- a/terraform/modules/g-drive-to-s3/10-lambda.tf
+++ b/terraform/modules/g-drive-to-s3/10-lambda.tf
@@ -110,11 +110,11 @@ resource "null_resource" "run_install_requirements" {
     dir_sha1 = sha1(join("", [for f in fileset(path.module, "../../../lambdas/g_drive_to_s3/*") : filesha1("${path.module}/${f}")]))
   }
 
-  provisioner "local-exec" {
-    interpreter = ["bash", "-c"]
-    command     = "make install-requirements"
-    working_dir = "${path.module}/../../../lambdas/g_drive_to_s3/"
-  }
+#   provisioner "local-exec" {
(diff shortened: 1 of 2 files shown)


terraform/modules/g-drive-to-s3/10-lambda.tf
```diff
@@ -138,6 +138,7 @@ resource "aws_lambda_function" "g_drive_to_s3_copier_lambda" {
   s3_bucket        = var.lambda_artefact_storage_bucket
   s3_key           = aws_s3_object.g_drive_to_s3_copier_lambda.key
   source_code_hash = data.archive_file.lambda.output_base64sha256
+  layers           = ["arn:aws:lambda:eu-west-2:${data.aws_caller_identity.current.account_id}:layer:google-apis-layer:1"]
```
timburke-hackit, line 141: I was going to recommend referencing the resource from the module output, but it doesn't output the arns - I'll make a note to go back and do that at some point.

timburke-hackit (approved):

terraform/modules/g-drive-to-s3/10-lambda.tf
Tian-2017: Yeah, I've noticed that too. When I am creating a pull request for Marta's lambda, I will output the ARNs of all the layers

PR #1686, opened 2024-04-08
add permission and tidy up

Tian-2017:
I've addressed two tiny issues and spent some time tidying up the Terraform file, hoping to make it more user-friendly for our colleagues.
1. added `aws_lambda_permission `to invoke the Lambda function to resolve the issue Marta mentioned about CloudWatch not functioning as a trigger (a necessity for triggering Lambda functions).
2. changed the crawler name, which must be the same in both the Lambda and `aws_glue_crawler `block.
3. tidied up the massive permission block by using a nested map - this has made about 50 lines of code redundant.
4. attempted to add layer ARNs to the output and then use them. However, the issue is that in the ETL directory, there is no existing .tf file that references the `aws-lambda-layers` module, so I couldn't use the layer ARNs from its output. To simplify, I added a local variable.
5. did a little naming convention things, e.g. the lambda name because resource or module names are recommended to use snake_case.

I'm not 100% sure whether replacing "`arn:aws:lambda:eu-west-2:336392948345:layer:AWSSDKPandas-Python39:12`" with current account's layer called "`panas-2-1-4-layer`" is correct. I can't find the ID 336392948345 in any of my granted AWS account IDs. Could someone help me double-check?

Could you add your ideas or tweak the code directly to make it even more readable for other colleagues. Thank you.

### DIFF
--- a/terraform/etl/49-lambda-streetscene-ingestion.tf
+++ b/terraform/etl/49-lambda-streetscene-ingestion.tf
@@ -1,65 +1,82 @@
 locals {
   create_street_systems_resource_count = local.is_live_environment ? 1 : 0
-  traffic_counters_tables                = ["street-systems"]
-}
-
-data "aws_iam_policy_document" "streetscene_street_systems_raw_zone_access" {
-  statement {
-    actions = [
-      "s3:AbortMultipartUpload",
-      "s3:DescribeJob",
-      "s3:Get*",
-      "s3:List*",
-      "s3:PutObject",
-      "kms:ListAliases",
-      "kms:Encrypt",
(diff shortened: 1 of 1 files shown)


LBHMVillalobos (approved):
Looks good to me - thanks for the work

Tian-2017:
@timburke-hackit @LBH-wgreeff Hi Tim and Hi Wayne, may I have your review at your convenience? Thank you

timburke-hackit (approved):
looks good to me, neat.

LBH-wgreeff (approved):
LGTM

PR #1692, opened 2024-04-15
DPP-677 add linter to check changed python files

Tian-2017:
- A linter has been created using Flake8 to check the syntax of changed files. It will check every new push/commit and only affect the changed files, not the entire repository.
- The E501 error (line too long) has been ignored, as it only permits 80 characters. Everyone has a slightly different style of commenting, which is not a significant issue for readability.
- The linter takes approximately 15-20 seconds (fast)to check a large Python file, based on testing.
- Through the linter, we can identify common Python syntax errors in advance rather than after deployment. For example, an error might occur before "Error message: NameError: name 'current_date' is not defined". There are also suggestions to improve the quality of our code.

### DIFF
--- a/.github/workflows/flake8_linter_python_files.yml
+++ b/.github/workflows/flake8_linter_python_files.yml
@@ -1,0 +1,39 @@
+name: Lint Python Files - Flake8
+# flake8 to check the syntax of changed files but ingore E501 error (line too long)
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


Tian-2017:
Wayne suggested using pre-commit to display the linter outcome in the IDE as one of the standard approaches. However, since Tim mentioned that not all our repository users are using an IDE, we will keep the workflow in the repo.

timburke-hackit (approved):

LBH-wgreeff (commented):
LGTM

PR #1694, opened 2024-04-15
create a new layer AWSSDKPandas (aka awswrangler)

Tian-2017:
This early PR (https://github.com/LBHackney-IT/Data-Platform/pull/1686) replaced arn:aws:lambda:eu-west-2:336392948345:layer:AWSSDKPandas-Python39:12 with panas-2-1-4-layer, which caused some issues when Marta used it in Lambda.

arn:aws:lambda:eu-west-2:336392948345:layer:AWSSDKPandas-Python39:12 is an AWS built-in layer managed by AWS on account 336392948345.
We can install this AWSSDKPandas (also known as awswrangler - https://github.com/aws/aws-sdk-pandas) as our own layer using Wayne's Lambda layer module.

This new PR installs such a layer for use with Lambda.

### DIFF
--- a/lambdas/lambda_layers/requirements8.txt
+++ b/lambdas/lambda_layers/requirements8.txt
@@ -1,0 +1,1 @@
+awswrangler==3.7.2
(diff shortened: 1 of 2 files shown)


LBH-wgreeff (approved):
LGTM

PR #1698, opened 2024-04-16
grant department write to landing g-drive paths

timburke-hackit:
This will allow department users access to the g-drive path in landing zone so that they can manually upload files that were previously being retrieved from Google Drive. 

Needed to facilitate #1716


Tian-2017 (approved):

timburke-hackit:
no longer required

PR #1702, opened 2024-04-17
minor amendments

annajgibson:
- correct the s3 source paths for mmh notes and tenure tables
- remove name of crawler as default name takes precedence
- change crawler name in housing_mtfh_case_notes_enriched.py to default name

### DIFF
--- a/scripts/jobs/housing/housing_mtfh_case_notes_enriched.py
+++ b/scripts/jobs/housing/housing_mtfh_case_notes_enriched.py
@@ -102,7 +102,7 @@ def main():
     for words in word_lists:
         word_list = words[1]
         mmh_notes_df[f"flag_{words[0]}"] = mmh_notes_df['all_notes_cleaned'].apply(
-            lambda text: find_keyword(text, word_list))
+            lambda text, word_list=word_list: find_keyword(text, word_list))
 
     logger.info(mmh_notes_df.sample(10))
 


scripts/jobs/housing/housing_mtfh_case_notes_enriched.py
```diff
@@ -111,7 +111,7 @@ def main():
     mmh_notes_df = mmh_notes_df.drop(columns={'all_notes_no_punct', 'all_notes_no_punct_no_stop'})
     mmh_notes_df = add_import_time_columns_pandas(mmh_notes_df)
     mmh_notes_df.to_parquet(path=s3_output_location, partition_cols=PARTITION_KEYS)
-    glue_client.start_crawler(Name='Housing MTFH case notes enriched to refined')
+    glue_client.start_crawler(Name='housing-mtfh-case-notes-enriched-to-refined')
     logger.info(f'Refined casenotes written to {s3_output_location}')
```
Tian-2017, line 115: Thanks, Anna. I've double-checked the name in the log; it's your revised "`housing-mtfh-case-notes-enriched-to-refined`".

I've rechecked the source code of the module again - the crawler "name" is not an input name, so entering "`name = "${local.short_identifier_prefix}Housing MTFH case notes enrichment to refined`"" will be ignored by Terraform. Terraform doesn't throw an error for undeclared variables by default; it ignores them. Thus, the `name` specified earlier version is not working.

By default, it will use job_name_identifier = replace(lower(var.job_name), "/[^a-zA-Z0-9]+/", "-") as the crawler name, not allowing us to change it based on the documentation at https://github.com/LBHackney-IT/Data-Platform/blob/main/terraform/modules/aws-glue-job/03-inputs-derived.tf.

Now, all is good. Thanks for fixing it.

terraform/etl/53-aws-glue-housing-extract-mmh-vulnerabilities-to-refined.tf
```diff
@@ -24,17 +24,16 @@ module "housing_mtfh_case_notes_enriched_to_refined" {
     "--enable-glue-datacatalog"          = "true"
     "--enable-continuous-cloudwatch-log" = "true"
     "--additional-python-modules"        = "spacy==3.7.4"
-    "--s3_source_mtfh_notes"             = module.department_housing_data_source.raw_zone_catalog_database_name
+    "--s3_source_mtfh_notes"             = "s3://dataplatform-stg-raw-zone/housing/"
     "--source_table_mtfh_notes"          = "mtfh_notes"
-    "--s3_source_tenure"                 = module.department_housing_data_source.refined_zone_catalog_database_name
+    "--s3_source_tenure"                 = "s3://dataplatform-stg-refined-zone/housing/mtfh/"
```
Tian-2017, line 29: You are using a staging S3 bucket. We need this work in staging and prod, so need use below:
`s3://${module.raw_zone_data_source.bucket_id}/house/`
`s3://${module.refined_zone_data_source.bucket_id}/housing/mtfh`

scripts/jobs/housing/housing_mtfh_case_notes_enriched.py
timburke-hackit: the module outputs the crawler name, so you could have used that and imported it as an environment variable. 

don't think it's worth changing, but maybe helpful in the future :)

Tian-2017: Well spotted.

PR #1707, opened 2024-04-23
DPP-682 fix the compatibility issue of pandas layer

Tian-2017:
The current pandas layer has compatibility issue with other layers. After discussion with Marta, we've realised we need to use the the AWS self-managed pandas layer called `arn:aws:lambda:eu-west-2:336392948345:layer:AWSSDKPandas-Python39:12 `(AWSSDKPandas).

I've tried to install it using our lambda module but it's too large, so I reference the AWS self-managed pandas directly in the terrafrom file.

Some merged PR related to this
https://github.com/LBHackney-IT/Data-Platform/pull/1695
https://github.com/LBHackney-IT/Data-Platform/pull/1694
https://github.com/LBHackney-IT/Data-Platform/pull/1686

### DIFF
--- a/terraform/etl/49-lambda-streetscene-ingestion.tf
+++ b/terraform/etl/49-lambda-streetscene-ingestion.tf
@@ -124,9 +124,10 @@ module "street_systems_api_ingestion" {
   }
   layers = [
     # No where in the etl directory reference aws-lambda-layers module, so can't use layer ARNs from its output
-    # "strcontains" checks if the string layer contains the substring "arn:aws:". It returns true if the substring is found, and false otherwise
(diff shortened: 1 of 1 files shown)


Tian-2017:
Want to use strcontains function to match the string and return boolean, but our version of Terraform doesn't support it https://developer.hashicorp.com/terraform/language/functions/strcontains

timburke-hackit:
I think you can just use contains, no?

Tian-2017:
Now looks good. Could you help to do the code review? Thank you.
image

timburke-hackit (approved):
yup, looks like this works 👍

PR #1720, opened 2024-04-25
Remove CODEOWNERS (for now)

spikeheap:
Following on from a Slack conversation, this change removes the current `CODEOWNERS` file.

We're planning to roll out mandatory code review across all repositories in Hackney (there's some more detail
[here](https://docs.google.com/document/d/1lA0CLXRtYzEopoW6gywWa7n-eDIQTaxzIM4FMwUdQI8/edit). As part of this we plan to require review from code owners if they're present in Github. For this repo that would currently force @LBH-wgreeff  to approve all changes 😬.

Here's a proposal to remove the `CODEOWNERS` file, which would prevent that bottleneck without affecting your current workflow. 

### Consequences

- Wayne will no longer automatically receive notifications about PRs unless he's subscribed or watches the repo.
- When the global rules are introduced this repo wouldn't need an exception (and the global rules won't affect this repo as it already has the same requirements set).

### Alternative solutions
Another approach would be to update this `CODEOWNERS` file to remove the global match and only specify owners for portions of the codebase owned by the Data Platform team.

Using this approach any change that matches an owner would require review by them (and notify them by default). Where no pattern matches anyone in Hackney can review and approve the change.

This can always be introduced in a new `CODEOWNERS` file in future.


LBH-wgreeff (approved):
LGTM

PR #1727, opened 2024-04-26
Create fake_error.py

Tian-2017:
I am fake the error to test if I will receive email for commits on our Data-Platform


Tian-2017:
delete the test

PR #1731, opened 2024-04-29
Daroclark patch 10

daroclark:
Changes made to create new Lambda for former tenant transfer.  I am not sure if I need to add other bits for the Lambda to be created or whether it just reuses the "copy-from-s3-to-s3" that already exists?


daroclark:
Excellent, thanks Tim.

PR #1734, opened 2024-04-30
skip linting parking scripts

timburke-hackit:
The linter will flag many minor errors in scripts created using the Glue console, generally whitespace and unused imports.

Parking users do not use a local IDE for python development so are unlikely to rectify these themselves and as a result will receive failure notifications whenever they make minor changes to old scripts. 

Although this can be done at the script level by including the flag `# flake8: noqa`, this would disable any local linter for a user who was using an IDE with linting functionality who might want to make these corrections. 

This is an example output which fails the linting because of unused imports present in the boilerplate code and other formatting errors from the Glue Studio console created script:
image

After some quick changes and using the Black formatter the linter will still fail because of these whitespace errors that exist in the generated SQL string:
image


timburke-hackit:
superseded by #1772

PR #1735, opened 2024-04-30
dedupe resource names

timburke-hackit:
Removes additional conflicting resource names created by this module.

PR #1736, opened 2024-04-30
Add runtime variable

timburke-hackit:
The module hardcoded a deprecated version of nodejs. This PR allows an optional variable that keeps the existing Lambda as is but allows the new lambda to be deployed with the recommended version of nodejs.

Includes name dedupe changes from #1735


timburke-hackit:
This needs a provider bump in order to use any of the allowed nodejs versions.

timburke-hackit:
@daroclark looks like this will work, the terraform linter is happy at least so let's see if the plans come out alright

PR #1747, opened 2024-05-09
fix iam policy deplicate name s3-to-s3 lambda

Tian-2017:
we have this error"│ Error: creating IAM Policy (***-copy-from-s3-to-s3-lambda): EntityAlreadyExists: A policy called ***-copy-from-s3-to-s3-lambda already exists. Duplicate names are not allowed.", which shows up recently.

I am using a UUID to have a quick fix, it will avoid deplicated names. however, it will recreate a new policy each time changes are applied. so it's not the best solution, but it's a quick fix.

### DIFF
--- a/terraform/modules/copy-from-s3-to-s3/10-lambda.tf
+++ b/terraform/modules/copy-from-s3-to-s3/10-lambda.tf
@@ -84,10 +84,12 @@ data "aws_iam_policy_document" "copy_from_s3_to_s3_lambda" {
   }
 }
 
+resource "random_uuid" "lambda_uuid" {}
+
 resource "aws_iam_policy" "copy_from_s3_to_s3_lambda" {
   tags = var.tags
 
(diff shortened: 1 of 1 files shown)


timburke-hackit:
this is already fixed in #1735 but the reason this is in a bit of a mess is more complicated.

The root of this is in solving a couple of issues for creating a new lambda:

- The runtime environment used in the module is no longer supported by AWS, any of the available runtimes are not supported by the provider used in the module.
- Both of the Lambdas need to use the same role in order for the assume role action to work on the Rentsense side

I'm pretty sure there's something else but these should but these are the two big ones.  Anyway, it's all in hand, I just haven't been doing it as it's not in the current sprint / PoC work and Daro doesn't need it sorted urgently.

Tian-2017:
@timburke-hackit Sounds good. Thanks Tim. In this case, I am closing the PR. I saw that the latest merged deployment failed, so I created a quick PR. Sorry, I haven't read all the recent PRs.

PR #1751, opened 2024-05-10
remove copy ft module

timburke-hackit:
This module is failing to deploy because the node14.x runtime is no longer available and alternative run times are not supported by the provider. 

The underlying module will need reworking so removing this to unblock the core workflow.


Tian-2017 (approved):
thanks Tim

PR #1753, opened 2024-05-13
delete non current versions and imcomplete multipart uploads

Tian-2017:
Added `expire_noncurrent_objects_days `to delete the noncurrent versions of objects. The previous PR (https://github.com/LBHackney-IT/Data-Platform/pull/1750) created `expire_objects_days`, which only takes effect on the current version.

`abort_multipart_days `is for deleting some incomplete uploaded files, which are uncommon, but it is beneficial to have this variable here.

So, with these four variables, we can now completely manage the versioning and deletion of any S3 bucket.

### DIFF
--- a/terraform/core/10-aws-s3-buckets.tf
+++ b/terraform/core/10-aws-s3-buckets.tf
@@ -390,6 +390,8 @@ module "lambda_artefact_storage" {
   bucket_identifier = "dp-lambda-artefact-storage"
   versioning_enabled = false
   expire_objects_days = 60  
+  expire_noncurrent_objects_days = 30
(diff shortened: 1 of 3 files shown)


LBH-wgreeff (dismissed):
LGTM

Tian-2017:
Hey Wayne, I've made a new commit to stop deleting data in the lambda-artefact bucket. Now, it only disables versioning and will not delete data older than 60 days in this bucket for the current version. These data are not logs like the spark_ui bucket, we should always keep the latest version.
Could you approve again? Thanks @LBH-wgreeff

LBH-wgreeff (approved):
LGTM

PR #1757, opened 2024-05-14
create two buckets

Tian-2017:
- Create one bucket in staging and another in prod to test the CD process for ticket DPF-142.
- Both buckets can also be used later by MWAA to parse the files.

Hopefully these two bucket names have not taken by others.
Many thanks to Tim for sharing thoughts - very helpful

### DIFF
--- a/terraform/compliance/s3.feature
+++ b/terraform/compliance/s3.feature
@@ -11,6 +11,7 @@ Feature: S3
   @exclude_module.glue_scripts.aws_s3_bucket.bucket
   @exclude_module.glue_temp_storage.aws_s3_bucket.bucket
   @exclude_module.lambda_artefact_storage.aws_s3_bucket.bucket
+  @exclude_module.airflow.aws_s3_bucket.bucket
   @exclude_module.lambda_artefact_storage_for_api_account.aws_s3_bucket.bucket
   @exclude_module.landing_zone.aws_s3_bucket.bucket
   @exclude_module.liberator_data_storage.aws_s3_bucket.bucket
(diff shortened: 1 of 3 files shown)


timburke-hackit:
Guess the compliance rules have changed. It's failing on this module - I'll pick up in the morning.

https://github.com/LBHackney-IT/Data-Platform//blob/f3e877509df475dd1be920c11d4ef4ab98fd1b99/terraform/modules/dynamodb

timburke-hackit:
and this fail in the ETL workflow I think will be fine once core runs / deploys
│ Error: reading KMS Key (alias/***-s3-airflow): NotFoundException: Alias arn:aws:kms:eu-west-2:***:alias/***-s3-airflow is not found.

timburke-hackit:
> and this fail in the ETL workflow I think will be fine once core runs / deploys │ Error: reading KMS Key (alias/_**-s3-airflow): NotFoundException: Alias arn:aws:kms:eu-west-2:**_:alias/***-s3-airflow is not found.

On this - it's a poor implementation. These values should be output from the core state then imported using a data source to ETL.

timburke-hackit (approved):
This looks good to me. 

Will make a note somewhere to revisit the compliance - it doesn't really make sense to have to make an exception one way or the other to create a bucket

Tian-2017:
> This looks good to me.
> 
> Will make a note somewhere to revisit the compliance - it doesn't really make sense to have to make an exception one way or the other to create a bucket

Should I add an in-line comment to mention that the compliance  has been skipped for this bucket?

timburke-hackit:
oh sorry, I'd missed that you'd skipped both, I don't think you need the exclusion on L15 because that scenario passes:

image

timburke-hackit (approved):

Tian-2017:
Ah, u're right, thanks. I've revised it to keep the second one. @timburke-hackit

PR #1758, opened 2024-05-15
Daroclark patch 10

daroclark:
Glue Script and terraform for Rentsense former tenant


timburke-hackit:
I can have a bash at the SonarCloud issues if that's helpful?

daroclark:
> I can have a bash at the SonarCloud issues if that's helpful?

Yes please, I've satisfied the Linter but not sure what they relate to.

PR #1760, opened 2024-05-17
Patch mtfh reshape to refined

timburke-hackit:
Addresses a change in the underlying tables that has changed these values from structs to strings. 

I'll raise a separate PR so that if a table in the middle of the job fails the others will still be processed.


daroclark (approved):

Tian-2017 (approved):
looks good. Thank you

PR #1761, opened 2024-05-17
refactor reshape to refined

timburke-hackit:
Refactor of reshape to refined job:
- moves transformations into functions
- rework of transformations to simplify and reduce the number of variable assignments
- job will now continue to process tables even if an earlier table fails, raises failed tables and exceptions at the end


timburke-hackit:
RE: SonarCloud issues, glueContext here is boiler plate from AWS. I'll maybe come back and make the exception handling better.

timburke-hackit:
Closing this, I don't really remember why I started and would probably want to start fresh if I were to revisit it again

PR #1766, opened 2024-05-22
create a typo since permission to mwaa bucket fails

Tian-2017:


Tian-2017:
don't need to add  this $ sign

PR #1774, opened 2024-05-23
pin alpine version to 3.19

timburke-hackit:
Alpine 3.20 removed the aws-cli tool because of incompatibility with python version 3.12

See issue: https://github.com/alpinelinux/docker-alpine/issues/396


Tian-2017 (approved):
Thanks, Tim, for fixing.

timburke-hackit:
@Tian-2017 no problem, thanks for pointing me in the right direction

PR #1776, opened 2024-05-24
add MWAA requirements to source control

timburke-hackit:
Adds the requirements file to the repository. This is the same as was in s3 with the addition of the mysql provider.

Includes terraform to create the object in s3.


Tian-2017 (approved):
Thanks! This is looking good so far – wonderful to have a version control to avoid both of us edit the requirements.txt file. 

When we get our CD ticket on the new repo, it would be better to deploy the file to the S3 target folder through the new repo.

Typically, the DAG files, plugins, and requirements.txt should be in the same location. Also, in this way, we won't need to wait 15 minutes for the data platform linter and another 15 minutes for deployment.

timburke-hackit:
@Tian-2017 yeah, that makes sense. Easy enough to move it once that's set up 👍

PR #1777, opened 2024-05-24
remove quotes from job name reference

timburke-hackit:
fixes reference to the job name which is the output of the module - was referenced as a string


Tian-2017 (approved):

Tian-2017 (approved):
LGTM

PR #1781, opened 2024-06-04
Update parking_cycle_hangar_allocation_update 

timburke-hackit:
Replace paths with f-strings. This doesn't change functionality but is intended to trigger the workflow and update the deployed job script.

not sure why @barnesm707 changes from #1767 haven't been deployed


Tian-2017 (approved):
thanks

PR #1783, opened 2024-06-05
create airflow user in iam and store access key and secret key in secrete manager

Tian-2017:
Create an Airflow user in IAM and store the access key and secret key in the secret manager, using dataplatform and parking as examples. Only grant some permissions to dataplatform to use the Athena Operator (we can refine this later).

The format follows the Airflow connections format to allow Airflow to parse it from the secret manager (see the screenshot in the comment below)


Tian-2017:
image
image
These are the results

timburke-hackit:
the default region is the only blocking change. 

If you want to leave this as is though, it maybe makes sense to build out the children families services permissions and create that connection so that we can use it for this piece of work. Then we can build out the others later.

terraform/core/46-mwaa-deparments-conn-id-and-permission.tf
```diff
@@ -0,0 +1,125 @@
...
+          "logs:GetQueryResults",
+          "logs:DescribeLogGroups"
+        ],
+        Resource = "arn:aws:logs:*"
+      },
+      {
+        "Effect" : "Allow",
+        "Action" : [
+          "glue:CreateCrawler",
+          "glue:DeleteCrawler",
```
timburke-hackit, line 56: I'd maybe be tempted to remove this action to avoid the opportunity for someone to accidentally remove a crawler with a similar name to the one they actually wanted.

Admittedly, it's an edge case and probably only a theoretical risk anyway.

Tian-2017: @timburke-hackit  
Since we don't have cases that involve creating crawlers, updating data, and terminating the crawler, as I described a while ago when using a custom EMR operator to do so, we are going to manage the crawler through a Terraform file. Therefore, it would definitely fine to remove both the create and delete permissions.

terraform/modules/airflow-depermental-connections/02-inputs-optional.tf
```diff
@@ -0,0 +1,11 @@
+variable "region" {
+  description = "AWS region"
+  type        = string
+  default     = "us-west-2"
```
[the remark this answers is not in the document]
Tian-2017: Nice spot, thanks

PR #1788, opened 2024-06-06
adding egress and ingress rules for SMB share, port 445 and 139

LBH-wgreeff:
adding egress and ingress rules for SMB share, port 445 and 139


cgloker (approved):
✅

PR #1790, opened 2024-06-07
Update addressbasepremium_create_address_table.py

LBHSBALLEY:
Change 2 columns in output format to match addresses API table


LBHSBALLEY:
Wrong change

PR #1797, opened 2024-06-11
allow airflow dept user to pass role to dept glue agent

timburke-hackit:
The airflow department user needs to be able to pass the equivalent department Glue role so that it can invoke resources like Glue Crawlers.

This PR allows, for example, the housing airflow user to pass the housing glue role when invoking a Glue crawler.


Tian-2017 (approved):

timburke-hackit:
not actually needed - me getting confused by the glue crawler operator

PR #1805, opened 2024-06-17
Adding mwaa access to departments but will look to lock this down mor…

LBH-wgreeff:
Adding MWAA access to departments but will look to lock this down more after testing

### DIFF
--- a/terraform/modules/department/50-aws-iam-policies.tf
+++ b/terraform/modules/department/50-aws-iam-policies.tf
@@ -749,6 +749,24 @@ data "aws_iam_policy_document" "redshift_department_read_access" {
   }
 }
 
+// MWAA Access
+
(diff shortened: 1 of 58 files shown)


LBH-wgreeff:
issues with formatting

terraform/modules/department/50-aws-iam-policies.tf
```diff
@@ -749,6 +749,24 @@ data "aws_iam_policy_document" "redshift_department_read_access" {
   }
 }
 
+// MWAA Access
+
+data "aws_iam_policy_document" "mwaa_department_web_server_access" {
```
timburke-hackit, line 754: is this the only functional change?

That's a lot of formatting diffs!

PR #1808, opened 2024-06-20
change trigger job for parking_cycle_hangar_allocation_update

timburke-hackit:
This PR changes the preceding job that must run before _parking_cycle_hangar_allocaiton_update_

The job has been failing because it runs before the necessary tables are created.

### DIFF
--- a/terraform/etl/38-aws-glue-job-parking.tf
+++ b/terraform/etl/38-aws-glue-job-parking.tf
@@ -1408,7 +1408,7 @@ module "parking_cycle_hangar_allocation_update" {
   pydeequ_zip_key                = data.aws_s3_object.pydeequ.key
   spark_ui_output_storage_id     = module.spark_ui_output_storage_data_source.bucket_id
   script_name                    = "parking_cycle_hangar_allocation_update"
-  triggered_by_job               = "${local.short_identifier_prefix}parking_cycle_hangars_denormalisation_update"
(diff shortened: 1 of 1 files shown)


Tian-2017 (dismissed):

Tian-2017 (approved):
Not prob at all. Looks good, only one little concern: 
I think `local.is_live_environment` is always true if we deploy through the repo, based on how it is defined ([URL](https://github.com/LBHackney-IT/Data-Platform/blob/fab097bd31bf73ceeddba9c41366a97063957a3e/terraform/etl/03-input-derived.tf#L18)). When we deploy through GitHub action, it will use workspace "default".

You can directly use:
`module.parking_spreadsheet_parking_ops_cycle_hangar_list[0].crawler_name`

I may wrongly understand the automation, feel free to correct me

timburke-hackit:
> Not prob at all. Looks good, only one little concern: I think `local.is_live_environment` is always true if we deploy through the repo, based on how it is defined ([URL](https://github.com/LBHackney-IT/Data-Platform/blob/fab097bd31bf73ceeddba9c41366a97063957a3e/terraform/etl/03-input-derived.tf#L18)). When we deploy through GitHub action, it will use workspace "default".
> 
> You can directly use: `module.parking_spreadsheet_parking_ops_cycle_hangar_list[0].crawler_name`

prod and stg both use the default terraform workspace, dev everyone has their own workspace. This is effectively accounting for the fact the crawler resource doesn't exist in any of the dev workspaces because this module isn't deployed there:

https://github.com/LBHackney-IT/Data-Platform/blob/main/terraform/etl/08-google-sheets-imports.tf#L686

PR #1810, opened 2024-06-27
Update rentsense_former_tenants_to_refined.py

daroclark:
Job changed to include output to xml in gzip and other date reformatting.


Tian-2017 (approved):
Looks good

PR #1819, opened 2024-07-05
add passrole policy

Tian-2017:
> botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the RunTask operation: User: arn:aws:iam::120038763019:user/child-fam-services-airflow-user is not authorized to perform: iam:PassRole on resource: arn:aws:iam::120038763019:role/test_hackney_ecs because no identity-based policy allows the iam:PassRole action

A passrole error to run the ingestion in ecs, after I've manually added below permission (see screenshot) and test it again, the ingestion task works well now. No more issue. I am sorry for changing a few times.

image

### DIFF
--- a/terraform/modules/department/50-aws-iam-policies.tf
+++ b/terraform/modules/department/50-aws-iam-policies.tf
@@ -865,6 +865,17 @@ data "aws_iam_policy_document" "airflow_base_policy" {
       "arn:aws:ecs:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:cluster/*"
     ]
   }
+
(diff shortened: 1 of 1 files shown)


timburke-hackit (approved):
thanks for making these changes - I was going to suggest exactly this while we're still getting the other ECS bits sorted.

Tian-2017:
@timburke-hackit Thanks. I rushed to leave home to catch the train - I should have further refinements on Friday afternoon. Anyway, please feel free to submit commits or make any comments on my PRs if you think they're not right.

PR #1832, opened 2024-07-16
Update 24-aws-glue-tascomi-data.tf

LBHSBALLEY:
Change time of Sunday's ingestions as we currently seem to be operating during an API maintenece window. We are trying 10am instead of 10pm (Idox has not replied with a suitable time).


Tian-2017 (approved):
Thanks for fixing that! If no other downstream jobs will be impacted, it is ready to be merged.

PR #1839, opened 2024-07-30
correct script name value

timburke-hackit:
The script for the job `parking_nas_live_manual_updates_with_pcns_box23` was incorrect, this pr sets the value to the expected script.


Tian-2017:
@timburke-hackit @davidwinter may have to manually delete the existing Parquet files to avoid confusion

Tian-2017 (approved):
thanks

PR #1842, opened 2024-07-31
Update parking_cycle_hangar_allocation_update.py

barnesm707:
tidy code


timburke-hackit:
Hi @barnesm707, I'm going to close this as I included the same change in #1841 along with the updated trigger for the job

PR #1844, opened 2024-08-02
change default maintenance window

Tian-2017:
- The default maintenance is at 3:30 on Friday - change it to Sunday.

### DIFF
--- a/terraform/core/47-mwaa.tf
+++ b/terraform/core/47-mwaa.tf
@@ -213,7 +213,7 @@ resource "aws_mwaa_environment" "mwaa" {
   tags                  = module.tags.values
 
   airflow_configuration_options = {
-    "core.default_timezone"               = "utc"
+    "core.default_timezone"               = "Europe/London"
     "webserver.warn_deployment_exposure"  = "False"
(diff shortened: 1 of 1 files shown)


terraform/core/47-mwaa.tf
```diff
@@ -213,7 +213,7 @@ resource "aws_mwaa_environment" "mwaa" {
   tags                  = module.tags.values
 
   airflow_configuration_options = {
-    "core.default_timezone"               = "Europe/London"
+    "core.default_timezone"               = "utc"
```
timburke-hackit, line 216: will this change the displayed run time and affect scheduling in the UI?

could you not handle the utc time in the task?

Tian-2017: This does not impact the UI. It supports both local time and UTC, with two options available. Users can always click the button to adjust based on personal preference.
image

Yes, it will impact the existing DAGs since they use the Airflow system timezone.

Yes, I can. I am fine with not changing from London to UTC. Let me remove this part and only change the maintenance window.

timburke-hackit (approved):

PR #1846, opened 2024-08-06
remove unnecessary database qualifiers

timburke-hackit:


Tian-2017:
Approved. I think using a database qualifier is a good habit. However, since it has an environment variable ('prod') in the database, it causes analysts' extra work. We can avoid using the environment variable for new databases in the future.

Tian-2017 (approved):

PR #1864, opened 2024-08-23
change the vrm_ceo to vrm

Tian-2017:
There is an execution order issue: the `GROUP BY vrm_ceo` is done before the alias (rename vrm to vrm_ceo) (in SQL, group by will be executed earlier than SELECT), so it causes an error in Athena, even though it did not throw an error in Glue.


timburke-hackit (approved):

daclarkelbh (commented):
The Order by has not been changed to new name

group by vrm
order by vrm_ceo

daclarkelbh:
Not sure how to change the file myself

Tian-2017:
Ignore this - I've just explained to @daclarkelbh 

> The order by happens after SELECT, so it should use the vrm_ceo

PR #1873, opened 2024-09-04
Create-a-street-scene-airflow-user and remove the lambda

Tian-2017:
This is a joint effort with Marta, mainly to:
- Create an Airflow user (only required for the first time, not needed in the future).
- Delete Marta's old script in both Terraform and Python. Terraform is no longer needed, except for the small part that creates the Glue crawler (which has been moved to another file storing all the crawlers that Airflow uses). BTW, table_prefix is not needed, so I've removed it when I migrated


LBHMVillalobos (approved):
all good, discussed with Tian

PR #1891, opened 2024-09-12
add partitions to existing tables using insert into

timburke-hackit:
Uses the [INSERT INTO](https://docs.aws.amazon.com/athena/latest/ug/insert-into.html) method to add new partitions where a table already exists rather than dropping and rebuilding the table. 

Behaviour for new tables remains unchanged.


LBH-wgreeff (approved):
LGTM

PR #1917, opened 2024-10-01
Removing vpc peering connection

LBH-wgreeff:
This is to update our Terraform with the changes to remove the VPC peering connection.


Tian-2017 (approved):
Approved.
Is there any downstream work relying on this VPC peering connection?
Our glue crawler seems to have an issue—Marta mentioned it in the DAP channel, and it looks like Tim has checked it. Could you wait to merge this PR until that is sorted out?

PR #1925, opened 2024-10-07
add import date to partition and correct the single digit month and day too

Tian-2017:
This is the second half of ticket https://hackney.atlassian.net/browse/PDP-34

Add import date to partition and correct the single digit month and day too; there are noise which removing the white space.


LBH-wgreeff (approved):
LGTM

PR #1928, opened 2024-10-07
 edge case - correct Parking_Persistent_Evaders

Tian-2017:
it has hyphones in import_date like below and and single digit month and day.

parking/liberator/Parking_Persistent_Evaders/import_year=2022/import_month=3/import_day=5/import_date=2022-03-05


LBH-wgreeff (approved):
LGTM

PR #1932, opened 2024-10-09
add import_date as partition to Parking_Shop_Front_Licence_Totals

Tian-2017:
forgot to add import_date to partitionKeys, causing an error "The number of partition keys do not match the number of partition values"


LBH-wgreeff (approved):
LGTM

PR #1936, opened 2024-10-14
Feature/ce 2567 update qlik tags

acanbym:
This PR adds a new tag in order to facilitate accessing the Qlik servers. It does by adding a new, overriding 'Application=Qlik' tag on the primary Qlik(sense) server in Prod. 

Adds:
- tags to the local values in terraform/modules/qlik-sense-server/15-aws-ec2-prod.tf

Intended change as in the below screenshot:
image


timburke-hackit (approved):
This looks OK to me - we have noisy plans but there's nothing that looks out of the ordinary to me in these.

PR #1940, opened 2024-10-17
Add additional inputs including id fields and a dq dimensions map.

annajgibson:
Add assets_reshape table to dq testing.


tomc-hackney (approved):
Looks good

PR #1979, opened 2024-11-13
Dpf 220/ecs department roles

timburke-hackit:
Creates base department roles for running ECS tasks that are broadly equivalent to the existing Glue roles.


timburke-hackit:
Sorry about the mess of commits on this one. Was going to tidy up but will just squash and merge to keep main reasonably sensible.

timburke-hackit:
> Looks the ecs_department_policy miss the glue access, but feel free to merge, we can refine it later.

I figured they probably weren't needed because if you need an onward Glue job or crawler you'd likely be triggering that from Airflow. But yeah, if it turns out we need them it's easy enough to add another policy to do that. 👍

PR #1981, opened 2024-11-14
enable airflow user for env services department

Tian-2017:
Enable the Airflow user for department_environmental_services - this will create the departmental AWS connection in secrete manager.

Cliff wants to schedule one of his tasks under department_environmental_services, which primarily calls the Alloy API and then uses Alloy server to perform some work.


LBH-wgreeff (approved):
LGTM

PR #1990, opened 2024-11-21
Allow airflow user (parking and housing) one to pass the ecs-execution-role

Tian-2017:
> botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the RunTask operation: User: arn:aws:iam::120038763019:user/parking-airflow-user is not authorized to perform: iam:PassRole on resource: arn:aws:iam::120038763019:role/parking-ecs-execution-role because no identity-based policy allows the iam:PassRole action

Same as before, need to allow passrole permission here.


LBH-wgreeff (approved):
LGTM

PR #1993, opened 2024-11-22
Department ecs passrole

timburke-hackit:
The previous implementation would allow a department to pass the ecs task role of another department. For example the finance department was able to pass the parking and housing ecs task roles:
image


timburke-hackit:
@Tian-2017 are these roles used in the mosaic ingestion or somewhere else?
`dap-ecs-task-role`
`dap-ecs-execution-role`

PR #1997, opened 2024-11-25
allow access to airflow env stg or prod in all containers

Tian-2017:
These keywords "stg" and "prod" are stored in the Secrets Manager under `airflow/variables/env` to allow access by containers.

The ARN of `airflow/variables/env` in 'prod' and 'stg' differs because AWS attaches a unique string to each. If there is a better way to manage access to airflow/variables/env, please let me know. Thanks

I know we can pass it from the Airflow DAG to the container, but reading it from Secrets Manager seems more straightforward


LBH-wgreeff (approved):
LGTM

PR #1999, opened 2024-11-27
remove cloudwatch permission from task role

Tian-2017:
I am removing it from task role. The execution role created in Wayne's ecs module includes two permission: read from ECR and read/write to cloudwatch permission. The python script in containers can write to CloudWatch ECS group via it.

And this will fix the error LimitExceeded: Cannot exceed quota for PolicySize: 6144
https://github.com/LBHackney-IT/Data-Platform/actions/runs/12029136720/job/33534177911


LBH-wgreeff (approved):
LGTM

PR #2028, opened 2024-12-16
Update 24-aws-glue-spatial.tf

lcoulson2021:
amended epoch number for addressbase full load


LBHSBALLEY (approved):
Epoch name changed as planned.

PR #2042, opened 2025-01-07
DPF-226 remove crawlers from the source module and recreate it for each department

Tian-2017:
1. Move ingested Google Sheets to the Google Sheets folder, ensuring they are not scattered across the department folder in the raw zone.
2. Avoid the creation of crawlers from the source module.
3. Create three crawlers, one for each department (currently, only these three departments own Google Sheet ingestion Glue jobs).
4. Add temporary crawlers triggers that will be removed after enabling the airflow google sheet ingestion dag.


timburke-hackit (approved):

LBH-wgreeff (approved):
LGTM

PR #2056, opened 2025-01-13
upgrade airflow to 2.10.3

Tian-2017:
This is the latest version of MWAA based on the AWS doc below:
https://docs.aws.amazon.com/mwaa/latest/userguide/airflow-versions.html

The upgrade requires updating the Python package version to align with Airflow 2.10.3 and Python 3.11. This has been addressed in the following PR within the Airflow repo:
https://github.com/LBHackney-IT/dap-airflow/pull/270

The reason for upgrading the version of MWAA is that, from Airflow 2.10 onwards, there have been big changes to the UI, making it more user-friendly (see [URL](https://airflow.apache.org/blog/airflow-2.10.0/)).


LBH-wgreeff (dismissed):
LGTM

LBH-wgreeff (approved):

PR #2063, opened 2025-01-15
temp fix on the unrestricted_address_api_dbo_hackney_address

Tian-2017:
The address_api data has not been updated for the past seven days (see `s3://dataplatform-prod-raw-zone/unrestricted/addresses_api/dbo.hackney_address/import_year=2025/import_month=01/`).

Sandrine and I are not entirely sure which steps are causing the address_api to fail. We need to review the docs and possibly talk with other colleagues to identify the root causes.

For now, we’ve decided to extend the `push_down_predicate ` of 3 failed parking jobs to 30 days to allow the parking Glue jobs to function.

Looks a lot of changes - sorry it's formatting. Actually only changes the `push_down_predicate ` of table `unrestricted_address_api_dbo_hackney_address `to `30`

### DIFF
--- a/scripts/jobs/parking/parking_permit_denormalised_gds_street_llpg.py
+++ b/scripts/jobs/parking/parking_permit_denormalised_gds_street_llpg.py
@@ -1,17 +1,20 @@
-from calendar import c
 import sys
+from calendar import c
+
+from awsglue import DynamicFrame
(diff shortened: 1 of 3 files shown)


LBHSBALLEY (approved):
Looks good to me

PR #2064, opened 2025-01-15
Ce 2642 adding s3 bucket for mwaa etl scripts

LBH-wgreeff:
Adding new s3 bucket for ETL Scripts

### DIFF
--- a/terraform/compliance/s3.feature
+++ b/terraform/compliance/s3.feature
@@ -4,6 +4,7 @@ Feature: S3
   @exclude_module.qlik_server\[0\].aws_s3_bucket.qlik_alb_logs\[0\]
   @exclude_module.airflow.aws_s3_bucket.bucket
   @exclude_aws_s3_bucket.mwaa_bucket
+  @exclude_aws_s3_bucket.mwaa_etl_scrtipts_bucket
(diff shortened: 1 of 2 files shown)


timburke-hackit:
I'm inclined to go a step further and say the name should be service agnostic. The logical grouping is the task / workflow not the service the script is going to run on in my mind.

timburke-hackit:
I meant the bucket name - I can quite easily envisage a world where either:
- we move compute platform - for example to kubernetes, or if Airflow sort multi-tenancy environments in v3
- workflows that require or use more than one type of compute - so a combination of ECS tasks, Glue / EMR jobs etc

I think, to your point about not confusing users by how things are named, it's much easier to have a series of scripts that are in a structure similar to:

`etl_scripts/<department>/<workflow>/some_python_task.py`
`etl_scripts/<department>/<workflow>/some_spark_task.py`
`etl_scripts/<department>/<workflow>/sql/some_bit_of_sql.sql`
...

Rather than having to sync them all to different buckets. This is particularly true while people are manually uploading files to stg for testing.

LBH-wgreeff:
Just to add, the Python script, you can specify the bucket and an optional object within the buck, aka folder in the bucket.

PR #2065, opened 2025-01-16
Temporary fix for Glue jobs impacted by the Address API

Tian-2017:
There are two tables that need to be updated, which are causing the failure. In yesterday's [PR](https://github.com/LBHackney-IT/Data-Platform/pull/2063), we only updated one of them:

unrestricted_address_api_dbo_hackney_address
unrestricted_address_api_dbo_hackney_xref


LBHSBALLEY (approved):
Well spotted!

PR #2067, opened 2025-01-20
allow the airflow user to update and create partitions

Tian-2017:
`Exception: Query failed with state: FAILED, reason: Insufficient permissions to execute the query. User is not authorized to create partitions on Glue catalog . You may need to manually clean the data at location 's3://dataplatform-stg-athena-storage/parking/temp/tables/94ee04d0-5656-4a39-9c97-e37c0ad77aa4' before retrying. Athena will not delete data in your account.`

Attach permissions to the Airflow user to allow for partition creation and updates.


timburke-hackit:
The resource boundaries on this really need to be tightened up at some point.

timburke-hackit (approved):

PR #2068, opened 2025-01-20
Change maven command

timburke-hackit:
Changed command to `mvn assembly:single` to reflect main goal used in recent versions of maven assembly 
https://maven.apache.org/plugins/maven-assembly-plugin/
https://maven.apache.org/plugins/maven-assembly-plugin/single-mojo.html

I was trying to use `mvn package` but this created an empty jar. I think this is probably still the preferred way to build the jar, but will need a bit of time to reconfigure the pom.xml.


timburke-hackit:
@Tian-2017 not certain this will fix it, so let me know if you run into it not building again. 

The current configuration seems a bit messy to me looking at the docs, but there's not a lot of value in re-doing it if this works OK.

PR #2081, opened 2025-01-24
change two glue jobs' triggers

Tian-2017:
Mike has just raised that two glue tables have not been updated for a while.

I realised that the two glue jobs were relying on the Google Sheets module’s crawler. Previously, each sheet had its own crawler, but now all parking sheets share a single trigger.

Change the trigger to new crawler.

### DIFF
--- a/terraform/etl/38-aws-glue-job-parking.tf
+++ b/terraform/etl/38-aws-glue-job-parking.tf
@@ -1087,7 +1087,7 @@ module "parking_defect_met_fail" {
   pydeequ_zip_key                = data.aws_s3_object.pydeequ.key
   spark_ui_output_storage_id     = module.spark_ui_output_storage_data_source.bucket_id
   script_name                    = "parking_defect_met_fail"
-  triggered_by_job               = "${local.short_identifier_prefix}Copy parking Liberator landing zone to raw"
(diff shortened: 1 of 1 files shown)


terraform/etl/38-aws-glue-job-parking.tf
```diff
@@ -1087,7 +1087,7 @@ module "parking_defect_met_fail" {
   pydeequ_zip_key                = data.aws_s3_object.pydeequ.key
   spark_ui_output_storage_id     = module.spark_ui_output_storage_data_source.bucket_id
   script_name                    = "parking_defect_met_fail"
-  triggered_by_crawler           = local.is_live_environment ? module.parking_spreadsheet_parking_ops_db_defects_mgt[0].crawler_name : null
+  triggered_by_job               = "${local.short_identifier_prefix}Copy parking Liberator landing zone to raw"
```
timburke-hackit, line 1090: This doesn't have any relationship to the `parking_ops_db_defects_mgt` spreadsheet though - if the liberator dump didn't happen for whatever reason this job wouldn't run, but it looks like it's only dependent on the spreadsheet?

timburke-hackit: couldn't you trigger it the same as before but use the new crawler as the trigger?

timburke-hackit (approved):

PR #2093, opened 2025-01-28
Revert "removes HTTP and HTTPS traffic from coming in via the internet, as custodian removes these rules any way"

stonest:
Reverts LBHackney-IT/Data-Platform#2092


LBH-wgreeff (approved):
LGTM

PR #2108, opened 2025-02-05
create tascomi trusted zone in redshift and grant the data and insight user access

Tian-2017:
Daro wants the refined and trusted zone can be accessed by Qlik which is using the user "data_and_insight_department"


timburke-hackit (approved):
This looks fine to me. 

FYI, the scripting for how this works is here:
https://github.com/LBHackney-IT/Data-Platform/blob/b10888cd69cffed590e74ccf2ed0107be64f0dac/scripts/configure_redshift.py#L94

PR #2160, opened 2025-03-11
add read access to mtfh in landing  for d and i ecs tasks

timburke-hackit:
Allows read access to the MTFH exports in Landing for the Data and Insight ECS task role.

By design the landing zone is generally only accessible to data platform engineers. This data is needed for implementation of the housing data pipeline which is currently in development and will require access for other members of data and insight.


timburke-hackit:
This builds OK - I did a rebase after it had passed and I think the runners are just being flakey

PR #2195, opened 2025-03-27
create a crawler for the ingested synergy data

Tian-2017:
- Create a crawler for the ingested Synergy data
- add tags to all the crawlers in this file in the meantime.


timburke-hackit (approved):
thanks for adding the tags to the others too!

PR #2335, opened 2025-07-14
Update aws providers core

timburke-hackit:
Updates the AWS provider used in the `Core` state to `~> 5.0`. ETL and Networking will continue with current configuration using `~> 4.0`.

In order to make this change, many modules have had their constraints updated to `>= 4.0, < 6.0` ensuring they are compatible across Core, ETL and Networking. 

In `terraform/modules/kafka-event-streaming/50-cluster.tf` the `ebs_volume_size` variable has been removed and replaced with the equivalent value in a `storage_info` variable as it no longer exists.


LBH-wgreeff (approved):
LGTM

PR #2396, opened 2025-08-15
grant qlik user to access the files in the cloudtrail bucket

Tian-2017:
Got below error when building dashboard in the Qlik - needs to grant Qlik user permission to access the data
image


LBHMVillalobos (approved):
Thanks

PR #2440, opened 2025-09-10
add housing_nec_migration database to redshift schemas

timburke-hackit:
Adds the `housing_nec_migration` database to the redshift external schemas and grants the housing department access.


timburke-hackit:
@Tian-2017 
I also wasn't sure what the Qlik user was using - I guess D&I - I'll add them in, probably easiest for now.

PR #2746, opened 2026-05-15
Add reg leases data

annajgibson:


Tian-2017 (commented):
It looks like we need to add a cron time line based on the latest module. I think the playbook hasn’t been updated. I see most modules have
`ingestion_schedule = "cron(0 21 * * ? *)"`, 
you can change to `cron(0 2 ? * MON-FRI *) ` 02:00 UTC, Monday to Friday

Tian-2017 (approved):
Sorry, I’ve looked into this a bit further. This workflow is complex and includes Lambda and Glue jobs.

If no cron time is provided, it will use the default cron expression: cron(0 6 * * ? *), so that should be fine.

Lambda will still check the file-level metadata regularly using the default cron expression.

However, the downstream Glue job will not run the real ingestion if there are no changes to the CSV.

So it’s safe to merge.

PR #2755, opened 2026-05-21
Add staging-side access for dap-airflow dev branch S3 replication

Tian-2017:
Adds staging-side Terraform support for the dev-to-staging S3 replication path:

- Enables versioning on the staging MWAA destination buckets
- Adds lifecycle cleanup for noncurrent object versions
- Allows the dev replication role to write replicated objects into the staging buckets
- Allows the dev replication role to use the staging MWAA KMS key for replicated objects


Tian-2017:
not needed any more

PR #2768, opened 2026-06-15
Change the name of the database that contains Metastreet data

Tian-2017:
Daro spotted that the keyword Metastreet has been included in the table name, so it does not need to be included in the database name.

**private_sector_housing** will be the new name. If we have other databases from private sector housing in the future, we can load them into this catalog database.


stevefarrhackneygovuk (commented):
Approved.

stevefarrhackneygovuk (approved):
Approved.

PR #2787, opened 2026-07-27
Enable departmental CSV uploads to create Glue Catalog tables

Tian-2017:
## Summary

This PR enables six departments to upload CSV files through the AWS Console and automatically create corresponding Glue Catalog tables.

## Changes

- Uses `<department>/<target_table_name>/<file_name>.csv` as the required S3 structure.
- Creates the table name from `<target_table_name>` and ignores the CSV filename.
- Treats multiple CSV files in the same table folder as one table.
- Rejects CSV files with an invalid folder structure.
- Deletes the Glue table only after the final CSV in its folder is removed.
- Removes Unrestricted from the automation.
- Limits Lambda permissions to the enabled departments.
- Adds tests for table creation, path validation and deletion behaviour.

### DIFF
--- a/lambdas/csv_to_glue_catalog/main.py
+++ b/lambdas/csv_to_glue_catalog/main.py
@@ -8,8 +8,8 @@ Outputs:
 
 Operational notes:
     All generated columns use the Glue ``string`` type. The recommended S3 key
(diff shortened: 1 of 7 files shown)


stevefarrhackneygovuk (approved):
Approved.

PR #2789, opened 2026-07-27
Move SSO Glue permissions to managed policies and add policy size checks

Tian-2017:
### Error

> Error: putting SSO Permission Set (arn:aws:sso:::permissionSet/ssoins-68042240bf234e38/ps-cbaf72fa18f4109e) Inline Policy: operation error SSO Admin: PutInlinePolicyToPermissionSet, https response error StatusCode: 400, RequestID: 07b84b77-2d2e-47c1-b8b8-6b40b37e0f0c, ValidationException: Current size of the non-whitespace characters present in the InlinePolicy Document is 10405 bytes which has exceeded the maximum limit of 10240 bytes

### Changes
- Moves existing Glue permissions out of departmental SSO inline policies.
- Reuses the existing production Glue policy and creates a dedicated staging SSO Glue policy.
- Preserves the existing actions, resources, conditions and environment-specific permissions.
- Adds CI validation for the AWS 10,240 non-whitespace byte inline policy limit.
- Fails validation when an inline policy size is unknown.

### CI validation
image

### DIFF
--- a/.github/actions/.helpers/check-sso-inline-policy-size.sh
+++ b/.github/actions/.helpers/check-sso-inline-policy-size.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly plan_path="${1:-plan.out}"
+readonly maximum_size=10240
+
+failed=0
+
+policy_sizes=$(
+  terraform show -json "$plan_path" |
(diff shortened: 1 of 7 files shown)


stevefarrhackneygovuk (approved):
Approved.

PR #2790, opened 2026-07-28
Support quoted CSV and direct TSV uploads in Glue Catalog automation

Tian-2017:
## Summary

- Add automatic comma, Tab, and pipe delimiter detection for lowercase `.csv` uploads.
- Add direct support for lowercase .tsv uploads, using tabs as the delimiter. Tom needs this because he uses TSV files for spreadsheets; when exporting to CSV, text fields are not automatically enclosed in quotation marks.


stevefarrhackneygovuk (approved):
Approved.

PR #2817, opened 2026-09-03
Add sso permission set var

timburke-hackit:
Creates a new variable `create_sso_permission_set` in the department module to allow for permission sets to be removed without removing other resources or the associated Google Group. 

This new variable defaults to `false` so future instances of the department module will not create an SSO group by default. 

Removes SSO permission sets for 4 departments:
- Planning
- Customer Services
- HR and OD
- Streetscene

These groups are either not in use (identified by last use report from CE) or only contain members of D&I who can be granted access using the Lake Formation permission sets that supersede these groups. 

Planning includes one user who is not a member of D&I. I've spoken to them and they are not actively using the permissions set and they should be moved to a group in `dap-access-management` if they need access in the future.


stevefarrhackneygovuk (dismissed):
Approve

Tian-2017 (approved):

PR #2823, opened 2026-09-10
DL-274 Add global configuration for the max number of active tasks and runs per DAG.

Tian-2017:
- Add global configuration for the maximum number of active tasks and runs per DAG, as suggested by Cintia. This makes sense even though most active task limits are already specified at the individual DAG level.
- I've set MWAA defaults to **8** active tasks and **1** active run per DAG for stg and prod.

### DIFF
--- a/terraform/core/47-mwaa.tf
+++ b/terraform/core/47-mwaa.tf
@@ -250,6 +250,8 @@ resource "aws_mwaa_environment" "mwaa" {
 
   airflow_configuration_options = {
     "core.default_timezone"               = "Europe/London"
+    "core.max_active_tasks_per_dag"       = "8"
+    "core.max_active_runs_per_dag"        = "1"
     "webserver.warn_deployment_exposure"  = "False"
     "webserver.auto_refresh"              = "True"
     "scheduler.min_file_process_interval" = "180"


terraform/core/47-mwaa.tf
```diff
@@ -250,6 +250,8 @@ resource "aws_mwaa_environment" "mwaa" {
 
   airflow_configuration_options = {
     "core.default_timezone"               = "Europe/London"
+    "core.max_active_tasks_per_dag"       = "8"
```
timburke-hackit, line 253: This looks fine, but I don't think will actually do much to resolve the issue because the majority of tasks that are running on ECS are doing so with `deferrable = true` so won't take up an active task spot while the task is running on ECS and consuming an IP slot.

It looks like Pools might be able to manage this and include the deferred tasks:

https://airflow.apache.org/docs/apache-airflow/2.10.3/administration-and-deployment/pools.html

timburke-hackit (approved):

terraform/core/47-mwaa.tf
Tian-2017: Thanks, I agree. I don’t think this PR on its own addresses the underlying issue. I mainly made this change to align with Cintia’s recommendation and ease her concern.

I agree that a pool could help prevent IP exhaustion, especially if deferred tasks are included. However, our preference is not to further limit concurrent ECS tasks if we can avoid it — ideally, we’d increase the available IP capacity instead.

I’ll clarify this with Cintia and ask for her help in increasing the available IP capacity for our ECS/Fargate tasks. I’ve actually already pointed out quite clearly in the ticket that the issue is the lack of available IPs for ECS/Fargate tasks 
See my ticket https://hackney.haloitsm.com/ticket?id=292958

Tian-2017: Just added one more note to the Halo ticket to clarify the issue a bit further.

PR #2824, opened 2026-09-10
Create mosaic catalog databases and secrets

Tian-2017:
- Create two Mosaic catalogue databases via a new file `terraform/etl/61-aws-glue-catalog-database-lf-managed.tf,` 
- Create the Mosaic secret for storing Mosaic credentials.

PR #2825, opened 2026-09-11
DL-274 Add Mosaic ECS task role and MWAA PassRole permissions

Tian-2017:
Add a dedicated ECS task role to read the Mosaic secret and shared scripts, upload raw data, and use the required KMS keys. Allow MWAA to pass the Mosaic task and execution roles to ECS.

Stacked on #2824, which creates the Mosaic secret and catalog databases.


stevefarrhackneygovuk (approved):
Approve

